PR #27, opened 2015-10-27
Fix problems found from OOB testing

c1728p9:
Make the following changes:
-Add installer for the 'cat' command line tool
-Update location of mbed-driver library
-Update mbedhtrun command to reflect new name of test-hello.bin
-Add instructions on how to create test.txt


PrzemekWirkus:
Thank you!

Issue #45, opened 2015-12-17
greentea ignores import errors


bremoran:
This code:

```Python
try:
    import mbed_lstools
    import mbed_host_tests
except:
    pass
```

https://github.com/ARMmbed/greentea/blob/master/mbed_greentea/mbed_greentea_cli.py#L47

conceals import errors and makes the real cause obscure, which results in issues like this: 
https://github.com/ARMmbed/greentea/issues/44

ciarmcom:
ARM Internal Ref: IOTSYST-689

PrzemekWirkus:
Yes, this is a nasty one :) Gonna take care of that :)

PR #46, opened 2016-01-04
[Bugfix] greentea ignores import errors

PrzemekWirkus:
Fix for issue https://github.com/ARMmbed/greentea/issues/45.

Now dependency module ```xxxxxx``` import fails user will receive message like this:
```
mbedgt: Not all required Python modules were imported!
mbedgt: No module named xxxxxx
mbedgt: Check if:
        1. You've correctly installed dependency module using setup tools or pip:
                * python setup.py install
                * pip install <module-name>
        2. There are no errors preventing import in dependency modules
        See: https://github.com/ARMmbed/greentea#installing-greentea
```


PrzemekWirkus:
CI is failing because it can't import ```mock``` module used for testing.

0xc0170:
+1

Issue #39, opened 2015-12-01
Quickstart - CppUtest references using private module


0xc0170:
https://github.com/ARMmbed/greentea/blob/master/docs/QUICKSTART.md

As this is a private module, it's not available thus shall we remove the references for now? They will be replaced or point to a new module once it's ready.

PrzemekWirkus:
You can send PR! :)

ciarmcom:
ARM Internal Ref: IOTSYST-621

brianjang:
then you mean, for mbed OS 3.0 can't use cppUTest framework?
- as far as i know, mbed OS 3.0 has greentea test-framework.(ex, for testing mbed-drivers)
Could you give some information the below part?
-          How can I get cppUtest framework for mbed OS 3.0?
-          How can I build cppUtest test code in “yotta_modules\mbed-client-c\test”?
(from mbed github, mbed-client-c include "test" and "unittest" folder and these test code of these folder is using cppUTest framworks API)

PrzemekWirkus:
@brianjang 
I think for general docs you need to use: https://github.com/ARMmbed/greentea/tree/master/docs
Just a note of clarification:
* Greentea is a program used to automate test execution on mbed devices. Basically it builds yotta module and tests inside and flashes all tests one by one to get test results.
* CppUTest is a test framework, C/C++ code used to build together with your test code e.g. to mock, perform assertions etc.
  * You can compile your modules with [```testDependency```](http://yottadocs.mbed.com/reference/module.html#testDependencies) to CppUTest module (not sure if published our module with CppUTest) 
  * [Example test dependencies](https://github.com/ARMmbed/core-util/blob/master/module.json#L24) for this yotta module can be updated with any yotta module with additional test frameworks like [CppUTest](http://cpputest.github.io) or [Unity](https://github.com/ThrowTheSwitch/Unity). As long it compiles! ;)

BTW:
>How can I build cppUtest test code in “yotta_modules\mbed-client-c\test”? (from mbed github, mbed-client-c include "test" and "unittest" folder and these test code of these folder is using cppUTest framworks API)

I think they are using Unity there with ```mbed-client-c```. This solution is non canonical and you should file an Issue there (mbed-client-c) to ask them how to build stuff they have prepared with their module :)

niklas-arm:
[...]

PR #52, opened 2016-01-06
Updates repated to supported Unity/CppUTest frameworks

PrzemekWirkus:


0xc0170:
Shall we add utest reference? It provides greentea handlers so it's greentea friendly :-)

PrzemekWirkus:
Yes, we will provide Unity (utest?) reference, how about we talk about this.
We can add it here: https://github.com/ARMmbed/greentea/issues/39 to solve this issue.

0xc0170:
> Yes, we will provide Unity (utest?) reference, how about we talk about this.

Yes, cpputest is not public anyway, so this one can stay open if we consider cpputest to be useful with greentea.  I would like to see here utest, bringing @niklas-arm here.  
Not certain how descriptive we should be, but at least a pointer to utest module and its documentation?

docs/QUICKSTART.md
```diff
@@ -8,7 +8,8 @@ This document is a simple cookbook introducing the testing facilities available
 
 This document provides examples for all three methods of testing.
 
-* You can also create [unit tests](http://en.wikipedia.org/wiki/Unit_testing) with the [mbed compatible CppUTest library](https://github.com/ARMmbed/mbed-cpputest-private). Information about the CppUTest library can be found [here](https://cpputest.github.io/manual.html).
+* You can also create [unit tests](http://en.wikipedia.org/wiki/Unit_testing) with the mbed compatible CppUTest library. Information about the CppUTest library can be found [here](https://cpputest.github.io/manual.html) or 
+* use [Unity framework](https://github.com/ThrowTheSwitch/Unity) supported by mbed: [Unit](https://github.com/ARMmbed/unity) framework.
```
niklas-arm: Unit => Unit**y**

niklas-arm:
The `utest` Readme already contains an explicit example with greentea, so I would only point people there via a link and not mirror parts of the Readme here.

Our Unity fork only works with `utest`, as it has specific changes made so that the reporting works!

If people want to use Unity as a stand alone framework, they must use the official repo.

PrzemekWirkus:
I've mentioned utest and added link to example in documentation for ```greentea```.

docs/QUICKSTART.md
```diff
@@ -8,7 +8,9 @@ This document is a simple cookbook introducing the testing facilities available
 
 This document provides examples for all three methods of testing.
 
-* You can also create [unit tests](http://en.wikipedia.org/wiki/Unit_testing) with the [mbed compatible CppUTest library](https://github.com/ARMmbed/mbed-cpputest-private). Information about the CppUTest library can be found [here](https://cpputest.github.io/manual.html).
+* Use [Unity framework](https://github.com/ThrowTheSwitch/Unity) supported by mbed: [Unity](https://github.com/ARMmbed/unity) framework.
+  * Please have a look at our Unity integration under lean test hardness called [utest: A Simple C++ Test Harness](https://github.com/ARMmbed/utest). Utest documentation provides [examples how to use it with Greentea](https://github.com/ARMmbed/utest#example).
```
niklas-arm: "lean test hardness"?

PR #62, opened 2016-01-12
[Bugfix] Add more verboseness when telling user about missing mbed properties

PrzemekWirkus:
Add fix for issue https://github.com/ARMmbed/htrun/issues/51


Patater:
LGTM

Issue #48, opened 2016-01-05
Host test process is not being killed


0xc0170:
An example, udp server is not terminated when a test timeouts, this is observed under windows.

An example of the udp server:

```python
# Copyright 2015 ARM Limited
#
# Licensed under the Apache License, Version 2.0
# See LICENSE file for details.

import sys
import socket
import time
from sys import stdout
from SocketServer import BaseRequestHandler, UDPServer
from mbed_host_tests import BaseHostTest

class UDPEchoClient_Handler(BaseRequestHandler):
    def handle(self):
        """ One handle per connection
        """
        data, socket = self.request
        if '{{end}}' in data:
            print
            print data
        else:
            stdout.write('HOST: {:>4} <= {}'.format(len(data), data))
        socket.sendto(data, self.client_address)

class UDPEchoClientTest(BaseHostTest):
    """
    """
    name = 'sockets_udp_echo_client'

    def send_server_ip_port(self, selftest, ip_address, port_no):
        c = selftest.mbed.serial_readline() # 'UDPCllient waiting for server IP and port...'
        if c is None:
            selftest.print_result(selftest.RESULT_IO_SERIAL)
            return
        selftest.notify(c.strip())

        # time.sleep(2)

        connection_str = ip_address + ":" + str(port_no) + "\n"
        selftest.notify("HOST: Sending server IP Address to target... {}".format(connection_str))
        selftest.mbed.serial_write(connection_str)

        c = selftest.mbed.serial_readline() # 'UDPCllient waiting for server IP and port...'
        if c is None:
            self.print_result(selftest.RESULT_IO_SERIAL)
            return
        selftest.notify(c.strip())
        return selftest.RESULT_PASSIVE

    def test(self, selftest):
        # We need to discover SERVEP_IP and set up SERVER_PORT
        # Note: Port 7 is Echo Protocol:
        #
        # Port number rationale:
        #
        # The Echo Protocol is a service in the Internet Protocol Suite defined
        # in RFC 862. It was originally proposed for testing and measurement
        # of round-trip times[citation needed] in IP networks.
        #
        # A host may connect to a server that supports the Echo Protocol using
        # the Transmission Control Protocol (TCP) or the User Datagram Protocol
        # (UDP) on the well-known port number 7. The server sends back an
        # identical copy of the data it received.
        SERVER_IP = str(socket.gethostbyname(socket.getfqdn()))
        SERVER_PORT = 32765

        # Returning none will suppress host test from printing success code
[...]

PR #63, opened 2016-01-12
Notify abort event to mbed-htrun via stdin pipe

mazimkhan:
Fix for issue #48 

For notifying abort event to mbedhtrun stdin pipe is opened while soawning. 
stdin is closed when mbedhtrun is required to be aborted. 
greentea waits for 5 seconds to mbedhtrun to cleanup and exit. If not it terminates mbedhtrun.

mbedhtrun is also updated in PR https://github.com/ARMmbed/htrun/pull/54 to read stdin and exit gracefully when stdin is closed.

### DIFF
--- a/mbed_greentea/mbed_test_api.py
+++ b/mbed_greentea/mbed_test_api.py
@@ -182,8 +182,10 @@ def run_host_test(image_path,
                 pass
 
             # Give 5 sec for mbedhtrun to exit
+            ret_code = None
             for i in range(5):
                 ret_code = self.proc.poll()
+                # A None value indicates that the process hasn't terminated yet.
                 if ret_code is not None:
                     break
                 sleep(1)
@@ -194,7 +196,6 @@ def run_host_test(image_path,
                     self.proc.terminate()
                 except Exception as e:
                     print "ProcessObserver.stop(): %s" % str(e)
-                    pass
 
     def get_char_from_queue(obs):
         """ Get character from queue safe way
--- a/mbed_greentea/mbed_test_api.py
+++ b/mbed_greentea/mbed_test_api.py
@@ -19,7 +19,7 @@ Author: Przemyslaw Wirkus <Przemyslaw.wirkus@arm.com>
 
 import re
 import sys
-from time import time
+from time import time, sleep
 from Queue import Queue, Empty
 from threading import Thread
 from subprocess import call, Popen, PIPE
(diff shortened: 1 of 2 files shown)


mbed_greentea/mbed_test_api.py
```diff
@@ -174,12 +174,28 @@ def run(self):
...
-            except Exception as e:
-                print "ProcessObserver.stop(): %s" % str(e)
+                self.proc.stdin.close()
+            finally:
                 pass
 
+            # Give 5 sec for mbedhtrun to exit
+            for i in range(5):
+                ret_code = self.proc.poll()
+                if ret_code is not None:
```
PrzemekWirkus: Please add comment here:
```python
#  A None value indicates that the process hasn’t terminated yet.
```
so it would be easier to understand this code (why you just didn't write
```python
if ret_code:
   ...
```

mbed_greentea/mbed_test_api.py
```diff
@@ -174,12 +174,28 @@ def run(self):
...
+                    break
+                sleep(1)
+
+            if ret_code is None:            # Kill it
+                print 'Terminating mbed-host-test(mbedhtrun) process (PID %s)' % self.proc.pid
+                try:
+                    self.proc.terminate()
+                except Exception as e:
+                    print "ProcessObserver.stop(): %s" % str(e)
+                    pass
```
PrzemekWirkus: Dead code.

mbed_greentea/mbed_test_api.py
```diff
@@ -174,12 +174,28 @@ def run(self):
...
             try:
-                self.proc.terminate()
-            except Exception as e:
-                print "ProcessObserver.stop(): %s" % str(e)
+                self.proc.stdin.close()
+            finally:
                 pass
 
+            # Give 5 sec for mbedhtrun to exit
+            for i in range(5):
```
PrzemekWirkus: Can this loop look like this:
```python
# Give 5 sec for mbedhtrun to exit
for i in range(5):
    ret_code = self.proc.poll()
    if ret_code is not None:
        break
    sleep(1)
else:            # Kill it
    print 'Terminating mbed-host-test(mbedhtrun) process (PID %s)' % self.proc.pid
    try:
        self.proc.terminate()
    except Exception as e:
        print "ProcessObserver.stop(): %s" % str(e)
```

PrzemekWirkus:
+1

@mazimkhan This PR looks very good!

PrzemekWirkus:
@stegut01 Stefan, please also review this code.

0xc0170:
I pulled this branch plus the PR in htrun referenced here, tested with tcp server, the socket is closed and I can run the test multiple times.

PrzemekWirkus:
@0xc0170 Sounds good, I think Azim will refactor a little this PR and we can merge + release new greentea+htrun tandem :)

PrzemekWirkus:
Plus this PR breaks CI :P It has to be fixed also before we merge.

mbed_greentea/mbed_test_api.py
mazimkhan: This would be bit confusing. But I realise that ret_code should be declared outside for loop.

mazimkhan:
@PrzemekWirkus @stegut01 comments incorporated. Please give a final review if you feel.

Issue #66, opened 2016-01-15
[Concept] Add support for test case id tagging


PrzemekWirkus:
# Description
One binary can contain may test cases (test fixtures) with a unique test case ids.
We could add support to yield many test case id results per one test case execution.

## Related PRs and changes:
* https://github.com/ARMmbed/greentea/pull/67 (Added reporting, still JUnit reporting is not added)
* https://github.com/ARMmbed/htrun/pull/58 (Added APIs and macros detection - test case result)
* https://github.com/ARMmbed/mbed-drivers/pull/141 (Added new test_env APIs to handle new macros)

## New success code for test case
New success code for test case is an integer with following properties:
Range of ```MBED_HOSTTEST_TESTCASE_FINISH()``` success:
```
+---------------+--------------+---------------+
|  success < 0  | success == 0 |  success > 0  |
+---------------+--------------+---------------+
| <------ ERROR |      OK      | FAIL -------> |
+---------------+--------------+---------------+
```

### Example simple test case inside current test source code
```c++
// Somewhere in C++ test case file in your yotta module...
// My test case TESTCASE_NAME implementation
{
    int success = 0; // 0 for success! ;)
    MBED_HOSTTEST_TESTCASE_START(TESTCASE_NAME);

    // Do your testing bit here...

    // Make sure to set "success" to proper value:
    // 0 - 'success' result of the test case
    // >0 - 'failure' result of the test case
    // <0 - 'inconclusive' result of the test case

    MBED_HOSTTEST_TESTCASE_FINISH(TESTCASE_NAME,success);
}
```
You can think about this also like this:
```c++
if (result_of_my_test_case == 0) {
    // OK case !
    success = 0;
} else if (result_of_my_test_case > 0) {
    // FAILed case ! 
    // You can add any arbitrary and meaningful for you value < 0.
    // E.g. you can do success = 404; to mark in your test case 
    // that "server" returned 404 error.
    success = 404;    
} else if (result_of_my_test_case < 0) {
    // ERROR, test case execution was inconclusive
    // You can add any arbitrary and meaningful for you value > 0.
    // E.g. success = -1; when your test framework cough exception.
    success = -1;
}
```

# Rationale
We need to move to new mode of work where one binary (test suite) can report multiple test cases results from automated run.

# Changes
* Added macros to ```mbed-drivers/test_env.h```:
```c++
#define MBED_HOSTTEST_TESTCASE_START(TESTCASE)  notify_testcase_start(#TESTCASE)
[...]

PR #67, opened 2016-01-18
 Add support for test case id tagging - implementing #66 

PrzemekWirkus:
** This is conceptual implementation, to understand better test case functionality **

This PR is an implementation part of  #66: "Add support for test case id tagging".
We want to add more detailed test case results from tests (binaries being run).

We want to move to official name of:
* *Test suite* for binary running set of test cases. TS name will be a binary name biult by yotta.
* *Test case* / *Test Case Id* is a part of TS and can return an success code: ```OK```, ```FAIL```, ```ERROR```.

### DIFF
--- a/mbed_greentea/mbed_test_api.py
+++ b/mbed_greentea/mbed_test_api.py
@@ -98,7 +98,7 @@ def run_host_test(image_path,
                   enum_host_tests_path=None,
                   run_app=None):
     """! This function runs host test supervisor (executes mbedhtrun) and checks output from host test process.
-    @return Tuple with test results, test output and test duration times
+    @return Tuple with test results, test output, test duration times and test case results
     @param image_path Path to binary file for flashing
     @param disk Currently mounted mbed-enabled devices disk (mount point)
     @param port Currently mounted mbed-enabled devices serial port (console)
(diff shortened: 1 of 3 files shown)


mbed_greentea/mbed_test_api.py
```diff
@@ -276,7 +276,11 @@ def get_auto_property_value(property_name, line):
...
 
-    result = None
+    RE_TESTCASE_START = re.compile("^\{\{(testcase_start);(\w+)\}")
+    RE_TESTCASE_FINISH = re.compile("^\{\{(testcase_finish);(\w+);(\d+)\}\}")
+
+    result = None           # Test suite result
+    result_test_cases = {}  # Test cases results
     update_once_flag = {}   # Stores flags checking if some auto-parameter was already set
     unknown_property_count = 0
     total_duration = 20     # This for flashing, reset and other serial port operations
```
stegut01: Do we need to update total_duration if we have multiple test cases?

PrzemekWirkus: Good point, I need to check that. Total duration should be for whole test suite execution time (for whole binary execution including flashing and reset).

PR #72, opened 2016-01-20
mbed platform commissioning steps for setting read/write permissions …

mazimkhan:
mbed platform commissioning steps for setting read/write permissions on serial device for all users.
With these steps mbedgt does not need to be executed with root/sudo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -674,3 +674,17 @@ Uninstalling mbed-greentea:
 Proceed (y/n)? Y
   Successfully uninstalled mbed-greentea
 ```
+
+# Commissioning mbed platforms (Linux)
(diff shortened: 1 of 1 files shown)


PrzemekWirkus:
@mazimkhan On the top there is a TOC (Table of Content), please link to 
```
# Commissioning mbed platforms (Linux)
```
there.

PrzemekWirkus:
To regenerate TOC you can use this tool: https://github.com/PrzemekWirkus/git-toc

```
$ pip install git-toc
```

PrzemekWirkus:
I will add it, it is a minor thing :)

PR #77, opened 2016-01-27
add arch_ble target from seeed

jacobrosenthal:
I found that tests were failing because greentea wasnt using the -combined.hex file with merged softdevice.
https://github.com/ARMmbed/mbed-drivers/issues/143

My seeed firmware reports target arch_ble which is an nrf51822 variant which is hardcoded here. 

It might be better to encode this information into the target instead?

### DIFF
--- a/mbed_greentea/mbed_target_info.py
+++ b/mbed_greentea/mbed_target_info.py
@@ -92,6 +92,20 @@ TARGET_INFO_MAPPING = {
                 "reset_method": "default",
                 "program_cycle_s": 4
             }
+        },
+    "ARCH_BLE" : {
+        "yotta_targets": [
+                {
+                    "yotta_target": "tinyble-gcc",
+                    "mbed_toolchain": "GCC_ARM"
+                }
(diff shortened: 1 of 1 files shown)


mbed_greentea/mbed_target_info.py
```diff
@@ -92,6 +92,20 @@
                 "reset_method": "default",
                 "program_cycle_s": 4
             }
+        },
+    "ARCH_BLE" : {
+        "yotta_targets": [
+                {
+                    "yotta_target": "tinyble-gcc",
```
0xc0170: target is called as the platform (tinyble) ? And I believe ARCH_BLE is the name coming from mbed-ls?

PrzemekWirkus: Yes!
See https://github.com/jacobrosenthal/target-tinyble-gcc/blob/master/target.json#L2 and https://github.com/jacobrosenthal/target-tinyble-gcc/blob/master/target.json#L21.

PR #89, opened 2016-03-03
Saving gcov .gcda files with relative path.

mazimkhan:
If mbed target is compiled with ```-fprofile-dir=.``` it will print .gcda file paths relative to ```./build/<yotta target>```. Changes here are to save these files with a prefix ```./build/<yotta target>```.


mazimkhan:
CI failure because corresponding mbed-host-test is not available in CI. error:
```No local packages or download links found for mbed-host-tests>=0.2.0```

PrzemekWirkus:
+1

PR #109, opened 2016-03-16
Check expression for user specified arguments in hook command

stegut01:
Because some test cases for now may not return any LCOV output, an empty test_case_name.info file is generated. This can not be merged to result.info and the automation fails. This PR fixes the issue by enabling a feature to check for empty files or if file exists with new specified arguments

User specified arguments are (so far):
* ```(...<<...>>):n``` -> remove content from hook command if file not exists
* ```(...<<...>>):e```-> remove content from hook command if file exists but empty
* ```(...<<...>>):i```-> remove content from hook command if file not exists OR if file exists but empty

where:
* ```(...)``` -> specify content to check and if necessary removed
* ```<<...>>``` -> specify part which is a path to a file

Example hooks.json:
```
{
    "hooks": {
        "hook_test_end": "",
        "hook_post_test_end": "$lcov --gcov-tool gcov (-a <<./build/{yotta_target_name}/test_case_1.info>>):i --output-file temp_result.info",
        "hook_post_all_test_end": "$lcov --gcov-tool gcov [(-a <<./build/{yotta_target_name}/{test_name_list}.info>>):i] --output-file result.info"
    }
}
```
parameters:
```
yotta_target_name = frdm-k64f-gcc
test_name_list = ['test_case_1', 'test_case_2', 'test_case_3']
```
resulting hook command for "hook_post_test_end" if ```test_case_1.info``` exists and is not empty:
```
$lcov --gcov-tool gcov -a ./build/frdm-k64f-gcc/test_case_1.info --output-file temp_result.info"
```
resulting hook command for "hook_post_all_test_end" if ```test_case_1.info```,  ```test_case_3.info``` exists and is not empty (```test_case_2.info``` is empty):
```
$lcov --gcov-tool gcov -a ./build/frdm-k64f-gcc/test_case_1.info -a ./build/frdm-k64f-gcc/test_case_3.info --output-file temp_result.info"
```

### DIFF
--- a/mbed_greentea/mbed_greentea_hooks.py
+++ b/mbed_greentea/mbed_greentea_hooks.py
@@ -13,6 +13,7 @@ limitations under the License.
 Author: Przemyslaw Wirkus <Przemyslaw.wirkus@arm.com>
 """
 
+import os
 import re
 import json
 from subprocess import Popen, PIPE
@@ -83,8 +84,8 @@ class GreenteaCliTestHook(GreenteaTestHook):
                 cmd = cmd_expand
                 if verbose:
                     gt_logger.gt_log_tab("hook expanded: %s"% cmd)
-
             cmd = cmd.format(**format)
+            cmd = GreenteaCliTestHook.check_parameters(cmd)
             if verbose:
                 gt_logger.gt_log_tab("hook formated: %s"% cmd)
         return cmd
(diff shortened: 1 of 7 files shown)


mazimkhan:
What is the reason for having options ```n```, ```e``` and ```i``` when in any of these cases we can't merge.

stegut01:
Regarding LCOV merging problem ```i``` option would be enough. But because hooks could also be used for other things I tried to make it as generic as possible. Options can be extended if needed for other use (not LCOV related).

mazimkhan:
If ```i``` is enough then that can be the implicit behaviour without having to specify any option. Less code:)

mbed_greentea/mbed_greentea_hooks.py
```diff
@@ -140,6 +141,57 @@ def expand_parameters(expr, expandables, delimiter=' '):
...
+            (...) -> specify part to check
+            <<...>> -> specify part which is a path to a file
+        'expr' expression (-a <<{build_path}/test/{test_name}.info>>):i will be either:
+            "-a ./build/frdm-k64f-gcc/test/test_name.info"
+            or will be removed from command
+        It is also possible to use it in combination with expand_parameters:
+        expr = "lcov --gcov-tool arm-none-eabi-gcov [(-a <<{build_path_list}/test/{test_name_list}.info>>):i] --output-file result.info"
+        """
+        result = expr
+        expr_strs_orig = re.findall('\(.*?\):[a-z]', expr)
```
PrzemekWirkus:
```python
expr_strs_orig = re.findall('\(.*?\):[nei]', expr)
```

PR #110, opened 2016-03-22
Add lcov hook

stegut01:
## Introduction
Because some test cases for now may not return any LCOV output, an empty test_case_name.info file is generated. This can not be merged to result.info and the automation fails. This PR fixes the issue by enabling a feature to check for empty files or if file exists.

## Usage
To enable code coverage use:
```
mbedgt -VS --hooks=lcov
```
By using a hook named ```lcov``` code coverage output is enabled and the files are autoamtically checked if they exists or empty.

### DIFF
--- a/mbed_greentea/mbed_greentea_hooks.py
+++ b/mbed_greentea/mbed_greentea_hooks.py
@@ -13,6 +13,7 @@ limitations under the License.
 Author: Przemyslaw Wirkus <Przemyslaw.wirkus@arm.com>
 """
 
+import os
 import re
 import json
 from subprocess import Popen, PIPE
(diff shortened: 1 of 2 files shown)


mbed_greentea/mbed_greentea_hooks.py
```diff
@@ -140,6 +141,71 @@ def expand_parameters(expr, expandables, delimiter=' '):
                 result = expr.replace(expr_str_orig, delimiter.join(expansion_result))
         return result
 
+class LcovHook(GreenteaCliTestHook):
+    """! Class used to define a LCOV hook
+    """
+    lcov_hooks = {
+        "hooks": {
+            "hook_test_end": "$lcov --gcov-tool gcov  --capture --directory ./build --output-file ./build/{yotta_target_name}/{test_name}.info",
+            "hook_post_test_end": "$echo lcov --gcov-tool gcov  --capture --directory ./build --output-file ./build/{yotta_target_name}/{test_name}.info",
```
PrzemekWirkus: We can remove this line with ```hook_post_test_end```, it was just for test :)

PrzemekWirkus:
+1

PrzemekWirkus:
Let's just remove one not needed hook with ```echo``` and we can merge and release this feature.

mazimkhan:
+1

PR #111, opened 2016-04-05
Bumping dependency for mbed-host-tests

bridadan:
# Problem
In the latest release of greentea, it passes the target id (TID) directly to mbedhtrun via the `-t` option. This wasn't added until v0.2.5 of `mbed-host-tests`, but the dependency in `setup.py` only specifies `mbed-host-tests>=0.2.0`. Therefore if I have a version of `mbed-host-tests` installed older than v0.2.5, it will just fail.

# Solution
Update the requirement in `setup.py` to be more specific

# Implementation
Updated the requirements in `setup.py` to be more specifc for `mbed-hosts-tests`. It now requires `mbed-host-tests>=0.2.5`.


PrzemekWirkus:
Rule for test tools is that latest from epoch (in this case 0.2.x) for each tool is compatible.
I will merge your PR because this is anyway a very good check for compatibility. Thanks for catching this!

jupe:
just wondering why you are not following semantic version numbering rules.. ?

bridadan:
Looks like this is caused because of the 0.x.x rules of semantic versioning.

@PrzemekWirkus Maybe it's time to make a full 1.x.x release?

PrzemekWirkus:
Yes, we can make a full semantic versioning release with 1.0.0 for all tools.
Also story of this tool versions is a long story about how we were forced to version in our <big project>.

PR #113, opened 2016-04-27
Making Greentea build agnostic by introducing a test spec input switch

mazimkhan:
Greentea originally only supports yotta artefacts. It assumes it is run inside a yotta module and gathers information from local file system. To make it generic for any other test artefacts we can support a test specification input. This specification can tell information like platform, tool chain, build artefacts path, test binaries, flash methods of test binaries etc.

Changes:
- New switch ```--test-spec``` that takes a json files argument. Contents format shown below.
- Moved yotta specific code and assumptions in mbed_yotta_api.py that now produces same test spec as shown below.

Test Spec example:
```
{
    "builds": {
        "K64F-GCC": {
            "platform": "K64F",
            "toolchain": "ARM_GCC",
            "base_path": "./build/frdm-k64f-gcc",
            "baud_rate": 9600,
            "tests": {
                "mbed-drivers-test-c_strings":{
                    "binaries":{
                        "usb": {
                            "flash_method": "usb",
                            "path": "./build/frdm-k64f-gcc/test/mbed-drivers-test-c_strings.bin"
                        }
                    }
                },
                "mbed-drivers-test-dev_null":{
                    "binaries":{
                        "usb": {
                            "flash_method": "usb",
                            "path": "./build/frdm-k64f-gcc/test/mbed-drivers-test-dev_null.bin"
                        }
                    }
                },
                "mbed-drivers-test-echo":{
                    "binaries":{
                        "usb": {
                            "flash_method": "usb",
                            "path": "./build/frdm-k64f-gcc/test/mbed-drivers-test-echo.bin"
                        }
                    }
                },
                "mbed-drivers-test-generic_tests":{
                    "binaries":{
                        "usb": {
                            "flash_method": "usb",
                            "path": "./build/frdm-k64f-gcc/test/mbed-drivers-test-generic_tests.bin"
                        }
                    }
                },
                "mbed-drivers-test-rtc":{
                    "binaries":{
                        "usb": {
                            "flash_method": "usb",
                            "path": "./build/frdm-k64f-gcc/test/mbed-drivers-test-rtc.bin"
                        }
                    }
                },
                "mbed-drivers-test-stl_features":{
                    "binaries":{
                        "usb": {
                            "flash_method": "usb",
                            "path": "./build/frdm-k64f-gcc/test/mbed-drivers-test-stl_features.bin"
                        }
                    }
                },
                "mbed-drivers-test-ticker":{
                    "binaries":{
                        "usb": {
                            "flash_method": "usb",
                            "path": "./build/frdm-k64f-gcc/test/mbed-drivers-test-ticker.bin"
                        }
                    }
                },
                "mbed-drivers-test-ticker_2":{
                    "binaries":{
                        "usb": {
                            "flash_method": "usb",
                            "path": "./build/frdm-k64f-gcc/test/mbed-drivers-test-ticker_2.bin"
                        }
                    }
                },
                "mbed-drivers-test-ticker_3":{
                    "binaries":{
                        "usb": {
                            "flash_method": "usb",
                            "path": "./build/frdm-k64f-gcc/test/mbed-drivers-test-ticker_3.bin"
                        }
                    }
                },
                "mbed-drivers-test-timeout":{
                    "binaries":{
                        "usb": {
                            "flash_method": "usb",
                            "path": "./build/frdm-k64f-gcc/test/mbed-drivers-test-timeout.bin"
                        }
                    }
                },
                "mbed-drivers-test-wait_us":{
                    "binaries":{
                        "usb": {
                            "flash_method": "usb",
                            "path": "./build/frdm-k64f-gcc/test/mbed-drivers-test-wait_us.bin"
                        }
                    }
                }
            }
        }
    }
}
```

Test spec entity relation diagram:

```
                                      -----------------------------
                                     |          Test Spec          |
                                      -----------------------------
                                                       1
                                                       |
                                                       |
                                                      m
                                      -----------------------------                        --------------------------------
                                     |          Test Build         | 1----------------- 1 |   Platform & Tool chain  |
                                      -----------------------------                        --------------------------------
                                                       1
                                                       |
                                                       |
                                                      m
                                      -----------------------------
                                     |              Tests             |
                                      -----------------------------
                                                       1
                                                       |
                                                       |
                                                      m
                                      -----------------------------
                                     |     Tests Binaries       |
                                      -----------------------------
```

Usage:
```
> mbedgt -V  --test-spec test_spec.json
```


bridadan:
Really great stuff @mazimkhan! I tried it locally and it worked it great.

### "Wall O' Comments"

1. The key underneath `builds`
  - Is this name arbitrary? Is there a specification for this? Or is there a convention we should follow?
    - One option: Keep name arbitrary, but by convention use `<platform>-<toolchain>`. So `K64F-GCC` would become `K64F-GCC_ARM` (`ARM_GCC` in this example should probably be  `GCC_ARM` as used in `mbed_target_info.py`, but now I'm nitpicking :smile: )
2. The name for `platform` hasn't always been well defined. We've used "target", "platform", and "board" interchangeably. @sg- any pointers here?
3. If a specific test case wants to use a different baud, then this is handled by the host test, right? All tests must at least do the setup step at the baud rate specified in the test spec, correct?
4. Anything significant about `base_path`? Is this where the test-spec file lives? Or does it mean something else? Or *will* it mean something else potentially in the future? I see the utility of it, just curious if this is being used currently.
5. I want to understand the `binaries` structure a little better. It looks like the two key pieces of information you're trying to capture here is:
  1. What is the path to this binary?
  2. How can this binary be loaded onto the target?
    - This is interesting because there are multiple ways to flash binaries and multiple types of binaries!
      - Types of binaries I can think of: Bootable (suitable for USB flashing) and Bootloadable (suitable for FOTA)
      - Types of flash methods: mbed Mass Storage Device, pyOCD flash_tool, External USB programmer, FOTA, other bootloader, etc
      - Transports for flashing: USB, Bluetooth, more in the future? (not my expertise here unfortunately... :P)
  - There is an argument for **not** specifying the flash method in the test-spec. This is because this file will be generated by a build system, which normally is not aware of **how** a target will be flashed.
  - The build system is aware of the **type** of binary it is producing though, like "bootable", "bootloadable", "debug", etc
  - I do see the need to know the flash method, but maybe this should not be specified in the test spec. Instead, the flash_method may be specified at runtime?
    - It could default to USB MSD flashing as this is the current behavior, but to override this you could specify `mbedgt ... --flash-method=ble_fota` or something like that.
  - So the `binaries` section might look something more like this:
  ```
  "binaries":{
    "bootable": {
      "path": "./build/frdm-k64f-gcc/test/mbed-drivers-test-c_strings.bin"
    }
  }
  ```
  - The flash method would know what "types" of binaries it can flash, so the default flasher would look for `bootable` under `binaries`. A Bluetooth FOTA flash method might look for `bootloadable` under `binaries`, etc.
  - If there is a case where the build system is 100% sure that the binary it produces only works with one type of flash method, then it could list the binary under a custom "type". So instead of `bootable` it could be `my_flash_method-bootable`, which would only be picked up by the custom flash method.

6. Love moving yotta specific stuff into its own file! +1
  - Since we are changing the default behavior of greentea, maybe we should include a warning if no built binaries are found in the case of a yotta project?
  - Just a reminder like `No built tests found. Please run "yotta build" first`.
  
Phew, sorry for the giant brain dump, there's a lot to consider here! Just want to emphasize that what you've done here is a big step in the right direction. Thanks for getting us moving on the right path!

mbed_greentea/mbed_target_info.py
```diff
@@ -297,3 +297,29 @@ def get_mbed_supported_test(mbed_test_case_name):
...
+    Gives platform property.
+
+    :param platform:
+    :return:
+    """
+    if platform in TARGET_INFO_MAPPING:
+        if property in TARGET_INFO_MAPPING[platform]['properties']:
+            return TARGET_INFO_MAPPING[platform]['properties'][property]
+        else:
+            raise Exception("Property %s not found in platform %s" % (property, platform))
```
PrzemekWirkus: I do not want if we can afford exception throw like this in CI.
I've noticed you are not catching those when calling ```get_platform_property()``` in other places.

mbed_greentea/tests_spec.py
```diff
@@ -0,0 +1,250 @@
...
+        return self.binaries
+
+    def get_binary(self, type):
+        """
+        Gives a test binary of specific flash type.
+
+        :param type:
+        :return:
+        """
+        return self.binaries[type]
```
PrzemekWirkus: This will throw ```KeyError```. Maybe just return like this:
```python
return binaries.get(type, None)
```
So we can check without try/catch if type is there.

mbed_greentea/tests_spec.py
```diff
@@ -0,0 +1,250 @@
...
+
+    def __init__(self, name, platform, toolchain, baud_rate, base_path):
+        """
+        ctor.
+
+        :param platform:
+        :param toolchain:
+        :return:
+        """
+        self.name = name
```
PrzemekWirkus: You have lots of these getters for these fields. You wanted to make them private and accesible via get_ only?
If so let's make it like this:
```python
        self.__name = name
        self.__platform = platform
        self.__toolchain = toolchain
        self.__baud_rate = baud_rate
        self.__base_path = base_path
```

PrzemekWirkus:
@bridadan 

> The key underneath builds Is this name arbitrary?

I'm not a specialist but it looks like it is. I can imagine people leave this ```name``` field empty or do nto include it and we are having default/canonical build name ```<platform>-<toolchain>```. Here Azim has to comment, I just looked at the code here.

> Is there a specification for this? 

Looks like this will not be a specification, rather a good rpactise to name build this or that but we can sort it out for sure.

> If a specific test case wants to use a different baud, then this is handled by the host test, right? All tests must at least do the setup step at the baud rate specified in the test spec, correct?

If a specific test case wants to use a different baud, then build system should rebuild whole build to compile in new baudrate. Baudrate is something we pass to build system and set while we compile serial "0" for stdio. Here baudrate is guidance for ```htrun``` to how open serial port to device when certain test in flashed.

> The name for platform hasn't always been well defined. We've used "target", "platform", and "board" interchangeably.

Yep, we can change how we call "device" but let's remember there is a ton of code in workspace_tools that use certain lingo already.

PrzemekWirkus:
@mazimkhan I've added minor comment to your code. Kinda too much OOP but I will survive ;)

PrzemekWirkus:
@mazimkhan We need to understand this below part better:
```json
"binaries":{
    "usb": {
        "flash_method": "usb",
        "path": "./build/frdm-k64f-gcc/test/mbed-drivers-test-c_strings.bin"
    }
}
```

Few comments, more generic for now:
* I would add to ```"builds"/<build_name>```` level ```"flash_method``` field so it would be propagate down for all binaries BUT each binary entry should be able to redefine ```"flash_method```. Less entries if ```"flash_method``` is common for all binaries (most cases probably).
Example:
```
{
    "builds": {
        "K64F-GCC": {
            "platform": "K64F",
            "toolchain": "ARM_GCC",
            "base_path": "./build/frdm-k64f-gcc",
            "baud_rate": 9600,
            "flash_method": "cp",
            "tests": {
                "mbed-drivers-test-c_strings":{
                    "binaries":{
                        "usb": {
                            "path": "./build/frdm-k64f-gcc/test/mbed-drivers-test-c_strings.bin"
                        }
                    }
                },
                "mbed-drivers-test-dev_null":{
                    "binaries":{
                        "usb": {
                            "flash_method": "fota_ble",
                            "path": "./build/frdm-k64f-gcc/test/mbed-drivers-test-dev_null.bin"
                        }
                    }
                },
...
```

* I would not call above ```usb``` key ```usb```. because now USb uis used for bunch of stuff from DAPlink's MSD to J-Link!
  * Maybe we can for ```flash_method``` method just use "copy_method" we how use for copy/flash ```htrun``` plugins? 
  * My point is, maybe it is better to just give lfash methods names istead of creating groups of flash methods like ```usb``` with subcategory ```some_fancy_flash_method```.
```
$ mbedhtrun --help
...
  -c COPY_METHOD, --copy=COPY_METHOD
                        Copy (flash the target) method selector. Plugin
                        support: copy, cp, default, eACommander, eACommander-
                        usb, shell, shutil, stlink, xcopy
...
```

My proposal (or rather one of the options) would be to make a test entry a list of dictionaries (like a list of simple structures with only ```path``` as mandatory (?) field. Example:
```
                    [
                        {
                            "path": "./build/frdm-k64f-gcc/test/mbed-drivers-test-c_strings.bin",
                            "flash_method": "cp",
                        },
                        {
                            "path": "./build/frdm-k64f-gcc/test/mbed-drivers-test-c_strings_fota.hex",
                            "flash_method": "fota_ble",
                            "md5": "992b54e05f60c16fd5dd6f18e5d31e60"
                        }
                    ]
```

Example of file.json:
```
{
    "builds": {
        "K64F-GCC": {
            "platform": "K64F",
            "toolchain": "ARM_GCC",
            "base_path": "./build/frdm-k64f-gcc",
            "baud_rate": 9600,
            "flash_method": "cp",
            "tests": {
                "mbed-drivers-test-c_strings":
                    [
                        {
                            "path": "./build/frdm-k64f-gcc/test/mbed-drivers-test-c_strings.bin",
                            "flash_method": "cp",
                        },
                        {
                            "path": "./build/frdm-k64f-gcc/test/mbed-drivers-test-c_strings_fota.hex",
                            "flash_method": "fota_ble",
                            "md5": "992b54e05f60c16fd5dd6f18e5d31e60"
                        }
                    ]
```

PrzemekWirkus:
@mazimkhan Regarding ```flash_method``` itself. JUst for clarification: In the past we were assuming that:
* One device can have more than one way of flashing it, but in most cases we will use mbed-enabled way (DAPlink's MSD feature).
  * But we never in practice used more than one way to flash and flash method was constant for all binaries per test run.
  * Example: Some devices have interface chip and can have FOTA feature (so we can flash them using BLE or MSD).
* For sure we will have more than binary per test, at least ```.elf``` and ```.bin```. For Nordic platforms all above + ```<name>.hex```, ```<name>_combined.hex```

PrzemekWirkus:
@bridadan Very good comments about flashing / booting etc. Different perspective. For me, like I said other process defines which methods for flashing we will actually use per test run. Maybe more than one for device, maybe not. Its for sure an important requirement.

I can imagine user story like this: _As a BLE device user I want to flash all tests in my test group <group_name> using DAPlink and using FOTA and get all results on test report.

mbed_greentea/mbed_greentea_cli.py
```diff
@@ -228,6 +230,10 @@ def main():
                     dest='hooks_json',
                     help='Load hooks used drive extra functionality')
 
+    parser.add_option('', '--test-spec',
+                    dest='tests_spec',
```
adbridge: Slight inconsistency in the naming here:
--test-spec
tests_spec 

Might be better to make them the same ?

mbed_greentea/mbed_greentea_cli.py
```diff
@@ -551,6 +547,18 @@ def main_cli(opts, args, gt_instance_uuid=None):
         print_version()
         return (0)
 
+    # Detect Source of test spec
+    # arguments or yotta api
+    if opts.tests_spec:
+        test_spec = TestSpec()
+        with open(opts.tests_spec, "r") as f:
```
adbridge: Should test that this file actually exists

PrzemekWirkus:
@bridadan @adbridge @mazimkhan 
So we still have pending things to discuss:
* Do we even consider a test campaign where we flash two different ways one device, e.g. copy + FOTA.
* Do we create class of binary types like you said @bridadan:
  * Something we just flash (usb-wire) using methods like copy, cp, j-link, 3rd party CLI tool
  * Something we flash using air-interface (FOTA) and will require some extra stuff like BLE tokens connected to host.
* Should we approach Greentea'a interface proposed here from build tool side and see what is available from build system anyway. Because we may expect a lot but build system is limited.
  * Example: Does build system actually know how we gonna flash, it only knows binaries it produces and their desired destination (MSD, FOTA, smoke signals, Morse code).
  * How we drive test scenario in real life with tools we have?

PrzemekWirkus:
@mazimkhan For now maybe we can ignore:
* Schemer for this interface, still to decide: data driven of code driven approach.
* Some code issues because we are prototyping here so I will not give coding styles/rules comments here anymore (well before we are ready with functionality).

mazimkhan:
@bridadan 

> The key underneath builds
> Is this name arbitrary? Is there a specification for this? Or is there a convention we should follow?
> One option: Keep name arbitrary, but by convention use <platform>-<toolchain>. So K64F-GCC would become K64F-GCC_ARM (ARM_GCC in this example should probably be GCC_ARM as used in mbed_target_info.py, but now I'm nitpicking :smile: )
 
Yes. This is something that will appear in greentea test summary as:
```
+--------------+---------------+-----------------------------+-----------+--------+--------+--------+--------------------+
| target       | platform_name | test suite                  | test case | passed | failed | result | elapsed_time (sec) |
+--------------+---------------+-----------------------------+-----------+--------+--------+--------+--------------------+
| K64F-GCC_ARM | K64F          | mbed-drivers-test-c_strings | c_strings | 1      | 0      | OK     | 8.7                |
| K64F-GCC_ARM | K64F          | mbed-drivers-test-dev_null  | dev_null  | 1      | 0      | OK     | 28.73              |
+--------------+---------------+-----------------------------+-----------+--------+--------+--------+--------------------+
```
```ARM_GCC``` is just data. It would be populated by the build system. I feel for execution purpose we don't need it. We will need some generic way of injecting certain parameters and keywords useful for the test report. That will go transparently in the report via JUnit.

> If a specific test case wants to use a different baud, then this is handled by the host test, right? All tests must at least do the setup step at the baud rate specified in the test spec, correct?

Yes a host test can dynamically change the baud rate. But also this is requirement for test specific meta data. At this point I don't know if that will come in this spec or handled transparently by greentea.

> Anything significant about base_path? Is this where the test-spec file lives? Or does it mean something else? Or will it mean something else potentially in the future? I see the utility of it, just curious if this is being used currently.

Actually this is used by greentea to find some coverage related files generated by target. Originally this path was constructed by greentea using yotta target as: ```build/<yotta target>/```.

> Point 5

I totally agree with you this part needs discussion. There are different thought on this:
* Perhaps for CI and automation we can stick with copying to the MSD device. As fota bin will require target to be already running a fota bin. 
* For any fixed target specific flash methods perhaps ```"properties"``` in ```mbed_target_info.py``` is right place. 
* Unless this info can be harnessed from the TARGET in workspace tools. Hence changes in build system. Example build system generates a ```.bin``` and a corresponding ```.info``` with meta data.

> Love moving yotta specific stuff into its own file! +1

I will test this if not already happening.

File: mbed_greentea/mbed_greentea_cli.py [hunk 6]
```diff
@@ -726,140 +645,116 @@ def main_cli(opts, args, gt_instance_uuid=None):
...
-                        break
-
-            # Configuration print mode:
-            if opts.verbose_test_configuration_only:
+    for test_build in test_spec.get_test_builds():
+        gt_logger.gt_log("processing target '%s' toolchain '%s' compatible platforms..." %
+                         (gt_logger.gt_bright(test_build.get_platform()),
+                                             gt_logger.gt_bright(test_build.get_toolchain())))
+
+        platform_name = test_build.get_platform()
```
adbridge: Minor point you are invoking test_build.get_platform() twice here , would be more efficient to move this line above the gt_logger line  and then use platform_name as the parameter?

mbed_greentea/mbed_greentea_cli.py
```diff
@@ -726,140 +645,116 @@ def main_cli(opts, args, gt_instance_uuid=None):
...
+        baudrate = test_build.get_baudrate()
+        gt_logger.gt_log("processing '%s' platform..."% gt_logger.gt_bright(platform_name))
```
adbridge: Haven't we just printed platform name above ?

mbed_greentea/mbed_greentea_cli.py
```diff
@@ -726,140 +645,116 @@ def main_cli(opts, args, gt_instance_uuid=None):
...
-                # Regression test mode:
-                # Building sources for given target and perform normal testing
-
-                yotta_result, yotta_ret = True, 0   # Skip build and assume 'yotta build' was successful
-                if opts.skip_yotta_build:
-                    gt_logger.gt_log("skipping calling yotta (specified with --skip-build option)")
+            if mbed_dev['platform_name'] == platform_name:
+                # We will force configuration specific baudrate
+                if mbed_dev['serial_port']:
+                    mbed_dev['serial_port'] = "%s:%d" % (mbed_dev['serial_port'], baudrate)
```
adbridge: what happens if this fails ?

mbed_greentea/mbed_greentea_cli.py
```diff
@@ -726,140 +645,116 @@ def main_cli(opts, args, gt_instance_uuid=None):
...
-                        test_exec_retcode += 1
-                    continue
-
-                # Regression test mode:
-                # Building sources for given target and perform normal testing
-
-                yotta_result, yotta_ret = True, 0   # Skip build and assume 'yotta build' was successful
-                if opts.skip_yotta_build:
-                    gt_logger.gt_log("skipping calling yotta (specified with --skip-build option)")
+            if mbed_dev['platform_name'] == platform_name:
```
adbridge: What happens if this fails ?

mbed_greentea/mbed_greentea_cli.py
```diff
@@ -726,140 +645,116 @@ def main_cli(opts, args, gt_instance_uuid=None):
...
+
+            for test_name in filtered_ctest_test_list_keys:
+                image_path = filtered_ctest_test_list[test_name].get_binary('usb').get_path()
+                test = {"test_bin":test_name, "image_path":image_path}
+                test_queue.put(test)
+
+            #for test_bin, image_path in filtered_ctest_test_list.iteritems():
+            #    test = {"test_bin":test_bin, "image_path":image_path}
+            #    test_queue.put(test)
+
```
adbridge: Shouldn't really have commented out code in submitted files.

mbed_greentea/mbed_greentea_cli.py
```diff
@@ -726,140 +645,116 @@ def main_cli(opts, args, gt_instance_uuid=None):
...
+            number_of_threads = 0
+            for mut in muts_to_test:
+                #################################################################
+                # Experimental, parallel test execution
+                #################################################################
+                if number_of_threads < parallel_test_exec:
+                    args = (test_result_queue, test_queue, opts, mut, build, build_path, greentea_hooks)
+                    t = Thread(target=run_test_thread, args=args)
+                    execute_threads.append(t)
```
adbridge: Can this fail ?

mbed_greentea/mbed_target_info.py
mazimkhan: The request for a ```property``` is made internally from greentea. It is not an external request. 
Question is what can be done if the property is not found. If nothing then ```Exception``` is right thing. At the moment this function has only one usage. That is for property ```"program_cycle_s"```. In the existing code I did not find any alternative to this hence an exception.

mbed_greentea/tests_spec.py
```diff
@@ -0,0 +1,250 @@
...
+    class for Test build.
+    """
+
+    def __init__(self, name, platform, toolchain, baud_rate, base_path):
+        """
+        ctor.
+
+        :param platform:
+        :param toolchain:
+        :return:
```
adbridge: Header needs updating

mazimkhan:
Updated test spec format:

```
{
    "builds": {
        "K64F-GCC": {
            "platform": "K64F",
            "toolchain": "ARM_GCC",
            "base_path": "./build/frdm-k64f-gcc",
            "baud_rate": 115200,
            "flash_method": "cp",
            "tests": {
                "mbed-drivers-test-c_strings":{
                    "binaries":[
                        {
                            "path": "./build/frdm-k64f-gcc/test/mbed-drivers-test-c_strings.bin"
                        }
                    ]
                },
                "mbed-drivers-test-dev_null":{
                    "binaries":[
                        {
                            "path": "./build/frdm-k64f-gcc/test/mbed-drivers-test-dev_null.bin"
                        }
                    ]
                },
                "mbed-drivers-test-echo":{
                    "binaries":[
                        {
                            "path": "./build/frdm-k64f-gcc/test/mbed-drivers-test-echo.bin"
                        }
                    ]
                },
                "mbed-drivers-test-generic_tests":{
                    "binaries":[
                        {
                            "flash_method": "cp",
                            "path": "./build/frdm-k64f-gcc/test/mbed-drivers-test-generic_tests.bin"
                        }
                    ]
                },
                "mbed-drivers-test-rtc":{
                    "binaries":[
                        {
                            "flash_method": "cp",
                            "path": "./build/frdm-k64f-gcc/test/mbed-drivers-test-rtc.bin"
                        }
                    ]
                },
                "mbed-drivers-test-stl_features":{
                    "binaries":[
                        {
                            "flash_method": "cp",
                            "path": "./build/frdm-k64f-gcc/test/mbed-drivers-test-stl_features.bin"
                        }
                    ]
                },
                "mbed-drivers-test-ticker":{
                    "binaries":[
                        {
                            "flash_method": "cp",
                            "path": "./build/frdm-k64f-gcc/test/mbed-drivers-test-ticker.bin"
                        }
                    ]
                },
                "mbed-drivers-test-ticker_2":{
                    "binaries":[
                        {
                            "flash_method": "cp",
                            "path": "./build/frdm-k64f-gcc/test/mbed-drivers-test-ticker_2.bin"
                        }
                    ]
                },
                "mbed-drivers-test-ticker_3":{
                    "binaries":[
                        {
                            "flash_method": "cp",
                            "path": "./build/frdm-k64f-gcc/test/mbed-drivers-test-ticker_3.bin"
                        }
                    ]
                },
                "mbed-drivers-test-timeout":{
                    "binaries":[
                        {
                            "flash_method": "cp",
                            "path": "./build/frdm-k64f-gcc/test/mbed-drivers-test-timeout.bin"
                        }
                    ]
                },
                "mbed-drivers-test-wait_us":{
                    "binaries":[
                        {
                            "flash_method": "cp",
                            "path": "./build/frdm-k64f-gcc/test/mbed-drivers-test-wait_us.bin"
                        }
                    ]
                }
            }
        }
    }
}

```

bridadan:
@mazimkhan Like we talked about during the sync today, since the build system isn't aware of the flash method, let's not include that in the test spec file.

Instead, the build system is aware of what type of binaries it is producing. So instead of `flash_method` under `binaries`, maybe we could put something like `binary_type`. And there could be a higher level `binary_type` under the test build to provide a default value that can be overridden deeper in the structure. Currently, we will probably only target bootable binaries that can be flashed over the mbed Mass Storage Device.

So the test spec would look like:

```
{
    "builds": {
        "K64F-GCC": {
            "platform": "K64F",
            "toolchain": "ARM_GCC",
            "base_path": "./build/frdm-k64f-gcc",
            "baud_rate": 115200,
            "binary_type": "bootable",
            "tests": {
                "mbed-drivers-test-c_strings":{
                    "binaries":[
                        {
                            "path": "./build/frdm-k64f-gcc/test/mbed-drivers-test-c_strings.bin"
                        }
                    ]
                },
                "mbed-drivers-test-dev_null":{
                    "binaries":[
                        {
                            "binary_type": "bootable",
                            "path": "./build/frdm-k64f-gcc/test/mbed-drivers-test-dev_null.bin"
                        }
                    ]
                },
                ...
            }
        }
    }
}
```

mbed_greentea/mbed_greentea_cli.py
mazimkhan: open does that for us. Example:
```
IOError: [Errno 2] No such file or directory: 'test_spec2.json'
```

mbed_greentea/mbed_greentea_cli.py
mazimkhan: Right. Its was a leftover print where we were first printing yotta target and then resolved platform name.

mbed_greentea/mbed_greentea_cli.py
mazimkhan: ```if mbed_dev['platform_name'] == platform_name:``` or ```if mbed_dev['serial_port']:```
First one is to find a suitable platform for testing this build. It wouldn't do ```muts_to_test.append``` if it fails.
Second, it should not be None. Hence removed that check. But as caution it would be good to add validation on mbedls output.

mbed_greentea/mbed_greentea_cli.py
mazimkhan: we will not select this device for this build.

mbed_greentea/mbed_greentea_cli.py
mazimkhan: No as it defaults to 1. And there is code at line 632 to validate that it is not < 1. ```number_of_threads``` is initialized to 0.

mazimkhan:
@bridadan  Yes sure that was discussed. I have just updated the PR. Now the format looks like:

```
{
    "builds": {
        "K64F-GCC": {
            "platform": "K64F",
            "toolchain": "ARM_GCC",
            "base_path": "./build/frdm-k64f-gcc",
            "baud_rate": 115200,
            "binary_type": "bootable",
            "tests": {
                "mbed-drivers-test-c_strings":{
                    "binaries":[
                        {
                            "path": "./build/frdm-k64f-gcc/test/mbed-drivers-test-c_strings.bin"
                        }
                    ]
                },
                "mbed-drivers-test-dev_null":{
                    "binaries":[
                        {
                            "path": "./build/frdm-k64f-gcc/test/mbed-drivers-test-dev_null.bin"
                        }
                    ]
                },
                "mbed-drivers-test-echo":{
                    "binaries":[
                        {
                            "path": "./build/frdm-k64f-gcc/test/mbed-drivers-test-echo.bin"
                        }
                    ]
                },
                "mbed-drivers-test-generic_tests":{
                    "binaries":[
                        {
                            "binary_type": "bootable",
                            "path": "./build/frdm-k64f-gcc/test/mbed-drivers-test-generic_tests.bin"
                        }
                    ]
                },
                "mbed-drivers-test-rtc":{
                    "binaries":[
                        {
                            "binary_type": "bootable",
                            "path": "./build/frdm-k64f-gcc/test/mbed-drivers-test-rtc.bin"
                        }
                    ]
                },
                "mbed-drivers-test-stl_features":{
                    "binaries":[
                        {
                            "binary_type": "bootable",
                            "path": "./build/frdm-k64f-gcc/test/mbed-drivers-test-stl_features.bin"
                        }
                    ]
                },
                "mbed-drivers-test-ticker":{
                    "binaries":[
                        {
                            "binary_type": "bootable",
                            "path": "./build/frdm-k64f-gcc/test/mbed-drivers-test-ticker.bin"
                        }
                    ]
                },
                "mbed-drivers-test-ticker_2":{
                    "binaries":[
                        {
                            "binary_type": "bootable",
                            "path": "./build/frdm-k64f-gcc/test/mbed-drivers-test-ticker_2.bin"
                        }
                    ]
                },
                "mbed-drivers-test-ticker_3":{
                    "binaries":[
                        {
                            "binary_type": "bootable",
                            "path": "./build/frdm-k64f-gcc/test/mbed-drivers-test-ticker_3.bin"
                        }
                    ]
                },
                "mbed-drivers-test-timeout":{
                    "binaries":[
                        {
                            "binary_type": "bootable",
                            "path": "./build/frdm-k64f-gcc/test/mbed-drivers-test-timeout.bin"
                        }
                    ]
                },
                "mbed-drivers-test-wait_us":{
                    "binaries":[
                        {
                            "binary_type": "bootable",
                            "path": "./build/frdm-k64f-gcc/test/mbed-drivers-test-wait_us.bin"
                        }
                    ]
                }
            }
        }
    }
}

```

mazimkhan:
@adbridge @PrzemekWirkus All comments incorporated. Please review again.

bridadan:
LGTM

PrzemekWirkus:
Please give ma a day to review this again... :)

jupe:
just some thoughts related flashing: from tester point of view I propably dont care how binary will be uploaded to target device, i just care that my tests can be executed. if I want to test FOTA I still need base binary to be flashed before I can even try FOTA methods.. but sure it depend on use scenario.

Issue #114, opened 2016-04-29
Test spec paths aren't relative to base_path


bridadan:
The recently introduce test spec has a field called `base_path`. It can be confusing because it seems like all binary paths would use this as the base path.

For instance, if `base_path: ./build/tests` and my tests are located at `./build/tests/test-case-*.bin`, then I would expect the `path` field for each test case to look like `./test-case-*.bin`. Currently, each test case still has to specify the full path: `./build/tests/test-case-*.bin`.

Furthermore, I would expect the `base_path` would be relative to the location of the test spec file. Instead, the path is relative to the current working directory.

ciarmcom:
ARM Internal Ref: IOTSFW-2513

PrzemekWirkus:
@mazimkhan Please have a look.

mazimkhan:
We have understand the reasons for these fields and take into account the environment where this code is running.
```base_path``` is used to indicate the path where build output is. It is introduced so that tools (greentea) using the test spec can generate artefacts in the same output dir as the binaries.
Yes the the binary ```path``` contains this path and the binary name. It is clear by the name chosen ```path```. 

If the name ```base_path``` is confusing we can rename it rather then putting sense to it by removing relative path from the binary ```path```. The proposal is good as it optimises binary path so that we only have to specify binary name there. But I prefer putting verbose information at the time of generating this spec over processing the path in Greentea. Because this processing will introduce another rule in the spec that image path is ```base_path```/```path```. We should keep it as dumb as possible.

Path is relative from the CWD where Greentea is launched. 
Taking test spec path as root path will introduce assumption that this file is in the root folder. It not wrong as it contains relative paths from root. But it sound circular. Generate test spec in root with paths relative to root -> take root from test spec path. What is wrong in implicit assumption that workspace root i.e. CWD of greentea is root.
Secondly, ```opts.test_spec``` will not be set in the case of Yotta  where we internally generate the test spec object tree.
Please see these comments for #115

bridadan:
Ah ok, the main purpose of `base_path` wasn't clear to me. If greentea is creating other artifacts during the test process then it makes sense to have a directory for this.

That being said, it wasn't clear to me that it is assumed that the paths in the test spec need to be relative to the working directory of greentea. The reason was I could specify a full relative or absolute path to the test spec, so I expected that the test spec was ran in the same way, regardless of where greentea was executed.

[...]

PR #115, opened 2016-04-29
Respect base_path for binary paths

bridadan:
This solves the issue described in #114.

If the base_path is not intended to have an effect on the binary paths, then the binary paths should at least be relative to the test_spec file, not the current working directory of greentea.


bridadan:
Closed due to explanation in #114

Issue #119, opened 2016-05-09
Greentea spawns multiple process for htrun equal to entries in test_spec.json


adbridge:
Running mbegt with the following test_spec:
```
{
    "builds": {
        "K64F-ARM": {
            "platform": "K64F",
            "toolchain": "ARM",
            "base_path": "./.build/K64F/ARM",
            "baud_rate": 9600,
            "tests": {
                "case_aynch_validate":{
                    "binaries":[
                        {
                            "binary_type": "bootable",
                            "path": "./.build/K64F/ARM/utest_porting.bin"
                        }
                    ]
                }

            }
        },
        "K64F-GCC": {
            "platform": "K64F",
            "toolchain": "GCC_ARM",
            "base_path": "./.build/K64F/GCC_ARM",
            "baud_rate": 9600,
            "tests": {
                "case_aynch_validate":{
                    "binaries":[
                        {
                            "binary_type": "bootable",
                            "path": "./.build/K64F/GCC_ARM/utest_porting.bin"
                        }
                    ]
                }

            }
        }

    }
}
```
This causes 2 htrun processes to be spawned:
```
        calling mbedhtrun: mbedhtrun -d D: -p COM15:9600:9600 -f "./.build/K64F/ARM/utest_porting.bin" -C 4 -c shell -m K64F -t 02400226a09a1e710000000000000000
000000005d43e3c9
        calling mbedhtrun: mbedhtrun -d D: -p COM15:9600:9600 -f "./.build/K64F/GCC_ARM/utest_porting.bin" -C 4 -c shell -m K64F -t 02400226a09a1e71000000000000
0000000000005d43e3c9
```
both of which try to access the hardware at the same time causing a windows error.

ciarmcom:
ARM Internal Ref: IOTSFW-2534

PrzemekWirkus:
Good one :D

PrzemekWirkus:
Let's use this issue to specify how test_spec::"builds"::[<KEY>::<ALUE>]* behaves.

bridadan:
I'd like formally specify the behavior of greentea when it comes to test specs.

# Changes

1. Add a `--test-builds` option, similar in function to the `--test-by-names` command. It accepts a comma separated list of test builds that correspond to test builds in a test spec file. It will run only these specified test builds.
2. Augment the reporter api to support multiple test build runs. For example, for JUnit reports, use the `<testsuite>` key to represent each test build.

# Process

[...]

PR #120, opened 2016-05-10
Patches on top of Greentea v0.2.11 (bug-fixes and allignment with test specification feature)

PrzemekWirkus:
## Description
Changes:
* Add filter by name for `TestSpec.get_test_builds()`
  * Polishing: Remove debug prints from [tests_spec](https://github.com/ARMmbed/greentea/blob/master/mbed_greentea/tests_spec.py)
* Bug-fix for #119
* Add correct prints to warnings (switch `-n`)
* `YottaModule` functionality polishing
* Add to `Junit.TestSuite.properties` values of `toolchain` and `target` so Viewer can take produce reporting.

```xml
<?xml version="1.0" ?>
<testsuites errors="0" failures="0" skipped="0" tests="4" time="0.300000190735">
	<testsuite errors="0" failures="0" name="unknown.frdm-k64f-gcc" skipped="0" tests="4" time="0.300000190735">
		<properties>
			<property name="toolchain" value="GCC_ARM"/>
			<property name="target" value="K64F"/>
		</properties>
```

* Error should be reported only if module.json exists but we were unable to load it for some reason

Example log with this issue: http://e108747.cambridge.arm.com:8080/job/mb/17/console
```
14:32:49 mbedgt: exporting to JUnit file 'JUnit_gcc_arm.xml'...
14:32:49 YottaModule: error -  [Errno 2] No such file or directory: '.\\module.json'
14:32:49 mbedgt: test suite report:
```

### DIFF
--- a/mbed_greentea/mbed_greentea_cli.py
+++ b/mbed_greentea/mbed_greentea_cli.py
@@ -382,7 +382,7 @@ def run_test_thread(test_result_queue, test_queue, opts, mut, build, build_path,
         # Some error in htrun, abort test execution
         if host_test_result < 0:
             break
-        
+
         single_test_result, single_test_output, single_testduration, single_timeout, result_test_cases, test_cases_summary = host_test_result
         test_result = single_test_result
 
@@ -833,7 +833,9 @@ def main_cli(opts, args, gt_instance_uuid=None):
         # Reports (to file)
         if opts.report_junit_file_name:
             gt_logger.gt_log("exporting to JUnit file '%s'..."% gt_logger.gt_bright(opts.report_junit_file_name))
-            junit_report = exporter_testcase_junit(test_report, test_suite_properties = get_test_suite_properties())
+            # This test specification will be used by JUnit exporter to populate TestSuite.properties (useful meta-data for Viewer)
+            junit_test_spec = test_spec if opts.test_spec else None
+            junit_report = exporter_testcase_junit(test_report, test_suite_properties = get_test_suite_properties(test_spec=junit_test_spec))
             with open(opts.report_junit_file_name, 'w') as f:
                 f.write(junit_report)
         if opts.report_text_file_name:
--- a/mbed_greentea/mbed_greentea_cli.py
+++ b/mbed_greentea/mbed_greentea_cli.py
@@ -148,7 +148,7 @@ def main():
 
     parser.add_option('-t', '--target',
                     dest='list_of_targets',
-                    help='You can specify list of yotta targets you want to build. Use comma to sepatate them.' +
+                    help='You can specify list of yotta targets you want to build. Use comma to separate them.' +
                          'Note: If --test-spec switch is defined this list becomes optional list of builds you want to filter in your test:' +
                          'Comma separated list of builds from test specification. Applicable if --test-spec switch is specified')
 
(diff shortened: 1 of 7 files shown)


PrzemekWirkus:
@bridadan @mazimkhan Please have a look.
This PR adds two properties to JUnit report under TestSuite.properties.
It is using test specification classes to fetch information about toolchain and target (platform).
We will in the future extend this so it [passes more to the reporter](https://github.com/mbedmicro/mbed/blob/master/workspace_tools/test_exporters.py#L292).

mbed_greentea/mbed_greentea_cli.py
```diff
@@ -833,7 +833,9 @@ def main_cli(opts, args, gt_instance_uuid=None):
         # Reports (to file)
         if opts.report_junit_file_name:
             gt_logger.gt_log("exporting to JUnit file '%s'..."% gt_logger.gt_bright(opts.report_junit_file_name))
-            junit_report = exporter_testcase_junit(test_report, test_suite_properties = get_test_suite_properties())
+            # This test specification will be used by JUnit exporter to populate TestSuite.properties (useful meta-data for Viewer)
+            junit_test_spec = test_spec if opts.test_spec else None
+            junit_report = exporter_testcase_junit(test_report, test_suite_properties = get_test_suite_properties(test_spec=junit_test_spec))
```
bridadan: Minor gripe: can you flatten this line out a bit? It's kinda hard to read.
```
junit_report = exporter_testcase_junit(test_report, test_suite_properties = get_test_suite_properties(test_spec=junit_test_spec))
```

Maybe something like this?
```
test_suite_properties = get_test_suite_properties(test_spec=junit_test_spec)
junit_report = exporter_testcase_junit(test_report, test_suite_properties = test_suite_properties)
```

bridadan:
I tried this locally and it works just fine. Other than my minor gripe mentioned above, looks good!

+1

mbed_greentea/mbed_greentea_cli.py
PrzemekWirkus: Yes, I refactored it before you've added a comment on it.

PrzemekWirkus:
@mazimkhan Please have a look at this PR, especially look at https://github.com/ARMmbed/greentea/pull/120/commits/3c40068e27bd6feef3ec1c0f38b3fecc8ebc3943 which changes the way we get build name from test spec.

PS: Guys, we agreed we will have one key:value pair in 'builds" for test spec for now so we can avoid:
* when running build system twice for different toolchain/target - merging few test specs into one.
* adding extra complexity to Greentea to handle multiple test sepcs - we may never use in CI this feature anyways. Just like we never used multiple targets for Greentea with `-t` switch.

bridadan:
@PrzemekWirkus If greentea only supports the running of one test spec now, then that's ok to reflect this in the reporting for the time being. Do we plan to support having multiple test builds in a test spec? It seems like greentea should either handle this (then we need to update the reporting) OR we need to choose now to say "only one test build per test spec".

@mazimkhan @adbridge

bridadan:
The broader discussion for the multiple test builds per test spec is occurring in this issue: https://github.com/ARMmbed/greentea/issues/119

This PR will be updated to reflect the outcome of that conversation.

PR #129, opened 2016-05-23
Fixing stack trace when a bad test name is specified.

bridadan:
If you specify an incorrect test name with the `-n` when using a test spec file, you get and `AttributeError` stack trace:

```
C:\m\proj>mbedgt -V --test-spec=test_spec.json -n bad-test-name
mbedgt: detecting connected mbed-enabled devices...
mbedgt: detected 1 device
        detected 'K64F' -> 'K64F[0]', console at 'COM238', mounted at 'D:', target id '0240000033514e450014500585d4004be981000097969900'
mbedgt: processing target 'K64F' toolchain 'GCC_ARM' compatible platforms...
mbedgt: using platform 'K64F' for test:
        target_id_mbed_htm = '0240000033514e450014500585d4004be981000097969900'
        daplink_hic_id = '97969900'
        mount_point = 'D:'
        daplink_automation_allowed = '0'
        daplink_daplink_mode = 'Interface'
        daplink_local_mods = '0'
        daplink_unique_id = '0240000033514e450014500585d4004be981000097969900'
        daplink_interface_crc = '0x33e2dd18'
        target_id = '0240000033514e450014500585d4004be981000097969900'
        serial_port = 'COM238:9600'
        target_id_usb_id = '0240000033514e450014500585d4004be981000097969900'
        platform_name = 'K64F'
        platform_name_unique = 'K64F[0]'
        daplink_git_sha = '34182e2cce4ca99073443ef29fbcfaab9e18caec'
        daplink_interface_version = '0241'
        daplink_auto_reset = '0'
        daplink_usb_interfaces = 'MSD, CDC, HID'
        daplink_version = '0241'
mbedgt: test case filter (specified with -n option)
mbedgt: invalid test case names (specified with '-n' option)
mbedgt: test name 'bad-test-name' not found in 'test_spec.json' (specified with --test-spec option)
        note: test case names are case sensitive
        note: see list of available test cases below
mbedgt: unexpected error:
        'str' object has no attribute 'get_test_builds'
Traceback (most recent call last):
  File "C:\Python27\Scripts\mbedgt-script.py", line 9, in <module>
    load_entry_point('mbed-greentea', 'console_scripts', 'mbedgt')()
  File "c:\users\bridan01\documents\dev\greentea\mbed_greentea\mbed_greentea_cli.py", line 345, in main
    cli_ret = main_cli(opts, args)
[...]

PR #130, opened 2016-05-24
Add test_spec.json as default name for test spec

PrzemekWirkus:
## Description

Currently after we've added `--test-spec` switch users have to explicitly declare it after building with <build system> which outputs test spec. So:
* I've updated [README.md](https://github.com/ARMmbed/greentea/tree/devel_test_spec_pickup#test-specification-json-formatted-input) with some test specification info.
* if `test_spec.json` (which is default name of test spec returned by <build system>) exist in directory in which we are calling Greentea (`mbedgt` command) we will pick it up as default test spec. This action will be equal to calling Greentea with explicit `--test-spec test_spec.json` switch.
* If `module.json` exist in directory we will assume this is a yotta module and act accordingly.
* If both `module.json` and `test_spec.json` exist we will give priority to `test_spec.json` file and assume use wants to execute tests from default test specification.
* There is no change to `--test-spec <filename.json>` it still picks up test spec and executes tests from it.
* New functionality works with switches like `-V`, `-S` or `-l`/`--list`.
*  Bug-fix: We were loading a `None` (TestSpec.load(None)) to test_spec_filename  ctro parameter when `TestSpec()` ctor was called without parameters.

## Example
```
$ ls
test_spec.json
...
```
```
$ mbedgt -l
mbedgt: using 'test_spec.json' from current directory!
mbedgt: available tests for built 'K64F-GCC_ARM', location '.\.build/tests\K64F\GCC_ARM'
        test 'TESTS-integration-threaded_blinky'
        test 'TESTS-integration-basic'
```
This is equal to executing `mbedgt -l --test-spec test_spec.json`.

## Other
Note: `test_spec.json` will have higher priority than module.json file so if we are inside directory with module.json and test_spec.json we will use test spec file instead of using yotta's module.json file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -514,7 +514,7 @@ Above command will execute all tests for yotta module you are in, e.g. ```mbed-d
 
 # Test specification JSON formatted input
 
-Greentea originally only supports yotta artefacts. It assumes it is run inside a yotta module and gathers information from local file system. To make it generic for any other test artefacts we can support a test specification input. This specification can tell information like platform, tool chain, build artefacts path, test binaries, flash methods of test binaries to Greentea.
+Greentea originally only supports yotta artefacts. It assumes it is run inside a yotta module and gathers information from local file system. To make it generic for any other test artefacts we can support a test specification input. This specification can tell information like platform, toolchain, build artefacts path, test binaries, flash methods of test binaries to Greentea.
 Test specification is an interface which can be used by any build system or it can be created manually. Test specification interface was added to separate build system from test automation automation (Greentea).
 
 Changes:
@@ -523,11 +523,11 @@ Changes:
 
 ## Test specification formatted
 
-More detailed test specification format will be introduced in near future. In current form test specification is a dictionary with key-value pairs under "builds" entry where key is a build name and value is a dictionary with additional properties describing build itself. Build properties include platform name, toolchain used to compile interface chipa baudrate and list of test binaries.
+More detailed test specification format will be introduced in near future. In current form test specification is a dictionary with key-value pairs under "builds" entry where key is a build name and value is a dictionary with additional properties describing build itself. Build properties include platform name, toolchain used to compile, interface chip baudrate, and list of test binaries.
 
 ## Example of test specification file
 
-In below example there are two build defined:
+In the below example there are two builds defined:
 * Build `K64F-ARM` for Freescale `K64F` platform compiled with `ARMCC` compiler and
 * build `K64F-GCC` for Freescale `K64F` platform compiled with `GCC ARM` compiler.
 
@@ -584,7 +584,7 @@ In below examples we will use above test specification file.
 
 ### Command line usage
 
-When building your mbed projects with <build system> capable of returning test specification in our format you can directly call Greentea to execute tests or list available tests (`-l` / `--list` switch).
+When building your mbed projects with *build system* capable of returning test specification in our format you can directly call Greentea to execute tests or list available tests (`-l` / `--list` switch).
 
 #### Executing all tests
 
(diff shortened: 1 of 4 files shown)


PrzemekWirkus:
@bridadan @mazimkhan @adbridge Would you be so kind and review ?

README.md
```diff
@@ -506,6 +512,119 @@ $ mbedgt -V --target=frdm-k64f-gcc
...
+Test specification is an interface which can be used by any build system or it can be created manually. Test specification interface was added to separate build system from test automation automation (Greentea).
+
+Changes:
+* New command line switch `--test-spec` is introduced. It is used to pass test specification file name to Greentea.
+* Existing command line switch `-t` together with `--test-spec` switch can be used to select build(s) by name which should be used for test runs. When no test specification is defined switch `-t` / `--target` behaves as usual: you can select yotta targets inside yotta module.
+
+## Test specification formatted
+
+More detailed test specification format will be introduced in near future. In current form test specification is a dictionary with key-value pairs under "builds" entry where key is a build name and value is a dictionary with additional properties describing build itself. Build properties include platform name, toolchain used to compile interface chipa baudrate and list of test binaries.
```
bridadan: Nit-pick time!

> Build properties include platform name, toolchain used to compile interface chipa baudrate and list of test binaries.

to

> Build properties include platform name, toolchain used to compile**, interface chip baudrate,** and list of test binaries.

README.md
```diff
@@ -506,6 +512,119 @@ $ mbedgt -V --target=frdm-k64f-gcc
 
 Above command will execute all tests for yotta module you are in, e.g. ```mbed-drivers```.
 
+# Test specification JSON formatted input
+
+Greentea originally only supports yotta artefacts. It assumes it is run inside a yotta module and gathers information from local file system. To make it generic for any other test artefacts we can support a test specification input. This specification can tell information like platform, tool chain, build artefacts path, test binaries, flash methods of test binaries to Greentea.
```
bridadan: Nit-pick time!

> tool chain

to

> toolchain

(to be consistent with other instances of this word throughout the doc)

README.md
```diff
@@ -506,6 +512,119 @@ $ mbedgt -V --target=frdm-k64f-gcc
...
+* New command line switch `--test-spec` is introduced. It is used to pass test specification file name to Greentea.
+* Existing command line switch `-t` together with `--test-spec` switch can be used to select build(s) by name which should be used for test runs. When no test specification is defined switch `-t` / `--target` behaves as usual: you can select yotta targets inside yotta module.
+
+## Test specification formatted
+
+More detailed test specification format will be introduced in near future. In current form test specification is a dictionary with key-value pairs under "builds" entry where key is a build name and value is a dictionary with additional properties describing build itself. Build properties include platform name, toolchain used to compile interface chipa baudrate and list of test binaries.
+
+## Example of test specification file
+
+In below example there are two build defined:
```
bridadan: Nit-pick time!

> In below example there are two build defined:

to

> In **the** below example there are two **builds** defined:

README.md
```diff
@@ -506,6 +512,119 @@ $ mbedgt -V --target=frdm-k64f-gcc
...
+
+    }
+}
+```
+
+In below examples we will use above test specification file.
+
+### Command line usage
+
+When building your mbed projects with <build system> capable of returning test specification in our format you can directly call Greentea to execute tests or list available tests (`-l` / `--list` switch).
```
bridadan: `<build system>` doesn't show up in rendered Markdown, should this be something else?

README.md
```diff
@@ -506,6 +512,119 @@ $ mbedgt -V --target=frdm-k64f-gcc
...
+
+Assuming that `test_spec.json` is in current directory:
+```
+$ mbedgt -VS
+```
+will pick up test specification and execute all tests in it.
+
+#### Cherry-pick tests
+
+* We will first tests we want to execute:
```
bridadan: Nit-pick time!

> We will first tests we want to execute:

to

> We will first **list the** tests we want to execute:

bridadan:
I had some doc nitpicks, but everything else looks ok. I made a few comments on #129 about the fix made in this PR, but other than that things look good!

Issue #131, opened 2016-05-26
Remove dependency on mbed_target_info.py


bridadan:
Currently if any platform that is tested does not exist in `mbed_target_info.py`, then you get an error: `Platform <platform> not in target info store.`

This is only used when testing yotta modules, but it is not necessary when using a test spec file. In this case, greentea should make its best effort to test all platforms.

It looks the only info it's using from the target info store in the case of a test spec file is `program_cycle_s`. Ideally we would use a different mechanism for checking to see if the platform is ready. Though for the moment I think using a sensible default would be fine.

ciarmcom:
ARM Internal Ref: IOTSFW-2606

PrzemekWirkus:
I will handle this one.

PrzemekWirkus:
@bridadan Please have a look at #133 and review / test if it works for you.

bridadan:
Looks good!

PR #133, opened 2016-05-27
Add default behavior to 'get_platform_property'

PrzemekWirkus:
This is a response for #131

Changes:
* Add default behavior to 'get_platform_property'.
* Add `default` node to `TARGET_INFO_MAPPING` with default values across all platforms. Used when we call `get_platform_property`.


bridadan:
+1 LGTM

Issue #138, opened 2016-06-02
Incorrect meta data set in JUnit test report when mutliple test builds in test spec


bridadan:
If multiple test builds exist in a test spec, greentea will always select only one test build to set the `platform` and `toolchain` keys in the JUnit test report. This is best explained by example.

Lets say our test_spec contains two builds: `K64F-GCC_ARM` and `K64F-ARM` (where the build name corresponds to `<platform>-<toolchain>`). If greentea is executed, but only runs the `K64F-ARM` test build (via the `-t K64F-ARM` option), then the resulting JUnit test report will look like this:

```
<?xml version="1.0" ?>
<testsuites errors="0" failures="0" skipped="0" tests="52" time="72.5849995613">
	<testsuite errors="0" failures="0" name="K64F-GCC_ARM.K64F-ARM" skipped="0" tests="7" time="4.04999995232">
		<properties>
			<property name="toolchain" value="GCC_ARM"/>
			<property name="name" value="K64F-GCC_ARM"/>
			<property name="target" value="K64F"/>
		</properties>
		<testcase classname="K64F-GCC_ARM.K64F-IAR.mbed-os-frameworks-utest-tests-unit_tests-case_control_async" name="Control: Await" time="1.010000">
			<system-out>[1464841140.53][CONN][RXD] &gt;&gt;&gt; Running case #3: 'Control: Await'...
[1464841140.53][CONN][INF] found SYNC in stream: {{__sync;a583bdad-46ff-4553-967e-f01280dd2ef6}}, queued...
[1464841140.53][HTST][INF] sync KV found, uuid=a583bdad-46ff-4553-967e-f01280dd2ef6, timestamp=1464841140.534000
[1464841140.53][CONN][INF] found KV pair in stream: {{__version;1.1.0}}, queued...
[1464841140.53][HTST][INF] DUT greentea-client version: 1.1.0
[1464841140.53][CONN][INF] found KV pair in stream: {{__timeout;15}}, queued...
[1464841140.53][HTST][INF] setting timeout to: 15 sec
...
```

Notice that the `toolchain` and `name` properties are incorrect (as well as the `<testsuite>` *name* attribute). This is caused by the following line: https://github.com/ARMmbed/greentea/blob/master/mbed_greentea/mbed_yotta_api.py#L211

This was left over behavior from when having multiple test builds in a test spec was undefined. But now that we support this behvior, we need to properly support this case.

PrzemekWirkus:
[...]

PR #140, opened 2016-06-08
Fixing issues when multiple test builds exist

bridadan:
This PR addresses a few issues that arise when multiple test builds are present in a test spec.

## Changes
- Properly sets test build properties in the JUnit XML report. Fixes issue #138.
- Fixes issue where only one test build isexecuted even if no test build filter is applied.
- Removes yotta module name prefix from the test report
  - If this needs to be added back, then I'd propose adding the yotta module name to the "test build name", so it would look like `<yotta module name>.<yotta target name>`

@PrzemekWirkus please review
@mazimkhan, @adbridge, FYI

### DIFF
--- a/mbed_greentea/mbed_greentea_cli.py
+++ b/mbed_greentea/mbed_greentea_cli.py
@@ -28,6 +28,7 @@ from Queue import Queue
 from threading import Thread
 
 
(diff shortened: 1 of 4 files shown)


PrzemekWirkus:
+1

PR #144, opened 2016-06-15
Adding default forced_reset_timeout of 1 second

bridadan:
This is releated to https://github.com/ARMmbed/htrun/pull/99.

This provides a default forced_reset_timeout of 1 second in the target info script. Many targets aren't ready immediately after reset, so this adds a sensible delay before sending the sync command.


bridadan:
Ooops, wrong spot

PR #146, opened 2016-06-15
Bug-fix: program_cycles_s not loaded from targets.json

PrzemekWirkus:


PrzemekWirkus:
@bridadan @mazimkhan FYI

PrzemekWirkus:
@mlnx FYI... fix from your excel

PR #153, opened 2016-06-23
Fixed JSON formatted report - it should be stored to disk, not on screen (no way to catch output)

PrzemekWirkus:
Changes:
* Switch `--report-json <filename>` now dumps to text file instead of stdout.
  * Because other logs are captured on stdout we can't extract JSON anyway.
  * It will be used in the future to capture, push to `mbed.com` result database (of course not in the same format).
* Small code refactoring.

### Testing
```
$ mbedgt -VS --report-json __a.json --report-junit __a.xml --report-text __a.txt --parallel 4
```
```
...
mbedgt: exporting to JUNIT file '__a.xml'...
mbedgt: exporting to TEXT '__a.txt'...
mbedgt: exporting to JSON '__a.json'...
...
```

[__a.txt](https://github.com/ARMmbed/greentea/files/329967/__a.json.txt)

### DIFF
--- a/mbed_greentea/mbed_greentea_cli.py
+++ b/mbed_greentea/mbed_greentea_cli.py
@@ -295,10 +295,8 @@ def main():
                     help='You can log test suite results to text file')
 
     parser.add_option('', '--report-json',
-                    dest='report_json',
-                    default=False,
-                    action="store_true",
-                    help='Outputs test results in JSON')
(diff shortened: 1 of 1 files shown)


PrzemekWirkus:
@bridadan @mazimkhan @adbridge @ConorPKeegan FYI.

PR #177, opened 2016-08-08
Respects mbed-ls mocking.

bridadan:
mbed-ls pre version v1.1.4 used to always include mocked targets in the
listing. In v1.1.4, this moved to a separate function call
(mbedls_get_mocks). This commit calls this function and updates the
dependency on mbed-ls to >=v1.1.4


mbed_greentea/mbed_greentea_cli.py
```diff
@@ -696,6 +696,7 @@ def get_parallel_value(value):
 
     # Detect devices connected to system
     mbeds = mbed_lstools.create()
+    mbeds.mbedls_get_mocks()
```
mazimkhan: greentea should not invoke internal working of mbed-ls like. Perhaps mbed-ls should be changed to include mocks via create. Just as it does on command line.

bridadan: That makes sense to me. It was removed from the constructor of mbed-ls here: https://github.com/ARMmbed/mbed-ls/commit/1e48f7ad8386e50cf58307bb36748ba99185bc0c

Though it looks like it wasn't included back in to the flow when it was broken out into a separate function, so perhaps this was a mistake?

mazimkhan: I think it is a mistake. We have to fix it.

mazimkhan:
Fixed in Release https://github.com/ARMmbed/mbed-ls/releases/tag/v1.2.1

bridadan:
Thanks @mazimkhan ! I'll close this PR then.

Issue #182, opened 2016-08-25
html report file doesn't show details in a browser


cuevasj:
The attached (zipped) html file was generated using **mbedgt -V --report-html=my_report.html** (with the fix to mbed_report_api.py described in issue #181).

When I open the file (with Chrome, Firefox, or Internet Explorer) only OK or FAIL for each test
suite is shown.  None of the details like timing and test steps are shown, although they are actually
in the source html.  A screenshot of what I see is also attached.

It looks like clicking on OK or FAIL should result in some action, but when I do click them nothing happens.  As far as I can tell the JavaScript is not being executed.

[my_report.zip](https://github.com/ARMmbed/greentea/files/437428/my_report.zip)
![capture](https://cloud.githubusercontent.com/assets/19537551/17975839/3d06662c-6ab9-11e6-959b-d5f040873333.PNG)

bridadan:
@cuevasj Thanks for reporting this!

I was able to create an HTML report that let me view the complete log, though that was using the mbed OS 5 release. I did notice that the JavaScript was failing to find the proper element in the attached report you posted. I made a PR that should fix this issue: https://github.com/ARMmbed/greentea/pull/183

Would mind trying that PR and seeing if it fixes your issue? Or could provide the steps that you took to create that report so I can reproduce it?

Thanks!

cuevasj:
@bridadan You're welcome and thanks for fixing it!

Yes, the file from the PR generated a report that let me view the details - with the caveat that I had to apply the patch mentioned in #181.

For the record I'm running the mbed-drivers tests on a FRDM-K64F dev kit.  When our product is ready I'll switch to that.

bridadan:
Great, thanks @cuevasj! We'll look into the issue you mentioned in #181 and find a proper fix, I noticed even with the patch you described some of the information got jumbled.

@mazimkhan @ConorPKeegan This patch should be ready to go!

ciarmcom:
ARM Internal Ref: IOTSFW-2873

bridadan:
@cuevasj said https://github.com/ARMmbed/greentea/pull/183 fixed this issue for him, so this should be good to close @mazimkhan @adbridge

bridadan:
Bump, good to close @mazimkhan @adbridge

PR #183, opened 2016-08-25
Using getElementById for fetching elements in HTML report

bridadan:
This commit should fix an issue with the HTML report where sometimes
clicking on the OK/FAIL button does not bring up the test summary. It uses
the getElementById function in addition to the div id string to correctly
fetch the element and hide/show it.

This should help address #182.

Please review at @ConorPKeegan 
FYI @mazimkhan

### DIFF
--- a/mbed_greentea/mbed_report_api.py
+++ b/mbed_greentea/mbed_report_api.py
@@ -223,24 +223,30 @@ html_template = """
     <head>
         <title>mbed Greentea Results Report</title>
         <script type="text/javascript">
-            function toggleOverlay(elem) {
-                if (elem.style.display == "") {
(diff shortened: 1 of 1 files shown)


ConorPKeegan:
@bridadan Looks good to me

bridadan:
@ConorPKeegan Cool, thanks! Let's wait to hear from @cuevasj to see if it solves his problem, or see if he can provide details on how to reproduce the issue he saw so we can confirm that this fixes the issue.

bridadan:
Looks like @cuevasj said this fixed his issue in #182. Ready to go @mazimkhan

cuevasj:
Yes, this fixed my issue.  Sorry I didn't notice that you were waiting for me.

PR #184, opened 2016-09-05
Fix for Issue 181

### LINKED ISSUE #181: Trying to create html report results in error if platform or toolchain contain "-"
Using the command **mbedgt -V --report-html=my_report.html** resulted in the attached error.

I believe the root cause is that line 599 of mbed_report_api.py splits the string platform_toolchain without setting the optional maxsplit parameter to 1.  My toolchain is frdm-k64f-gcc so the result of splitting the string (on '-') is four substrings which cannot be assigned to two variables.

When I changed the line to **platform, toolchain = platform_toolchain.split('-', 1)**, the error goes away and the html is generated.  I am very new to Python so maybe there's a better way to fix this.

capture

ConorPKeegan:
# Changes
* When splitting up <PLATFORM>-<TOOLCHAIN>, we now only take the last element of the split (the toolchain, as it should not contain a '-'), and then remove that from the original string to get the platform
* Problem occurred when platform contained a '-'

# Fixes
* This pull request fixes #181

### DIFF
--- a/mbed_greentea/mbed_report_api.py
+++ b/mbed_greentea/mbed_report_api.py
@@ -598,7 +598,10 @@ def exporter_html(test_result_ext, test_suite_properties=None):
     # Populate a set of all of the unique tests
     for platform_toolchain, test_list in test_result_ext.iteritems():
         # Format of string is <PLATFORM>-<TOOLCHAIN>
-        platform, toolchain = platform_toolchain.split('-')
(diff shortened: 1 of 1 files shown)


ConorPKeegan:
@mazimkhan @bridadan Please can you review this fix

bridadan:
LGTM!

Issue #187, opened 2016-09-06
Multi-target test run chooses the incorrect target


bridadan:
When doing a multi-target test run, greentea will incorrectly call `mbedhtrun` with the only one of the target's information. For example:

```
$ mbedgt -V -v -t K64F-GCC_ARM,NUCLEO_F401RE-GCC_ARM
mbedgt: greentea test automation tool ver. 1.2.0
mbedgt: using multiple test specifications from current directory!
        using '.build\tests\K64F\ARM\test_spec.json'
        using '.build\tests\K64F\GCC_ARM\test_spec.json'
        using '.build\tests\KL25Z\GCC_ARM\test_spec.json'
        using '.build\tests\KL46Z\GCC_ARM\test_spec.json'
        using '.build\tests\NRF51_DK\GCC_ARM\test_spec.json'
        using '.build\tests\NUCLEO_F401RE\GCC_ARM\test_spec.json'
mbedgt: detecting connected mbed-enabled devices...
mbedgt: detected 2 devices
        +---------------+----------------------+-------------+-------------+--------------------------------------------------+
        | platform_name | platform_name_unique | serial_port | mount_point | target_id                                        |
        +---------------+----------------------+-------------+-------------+--------------------------------------------------+
        | NUCLEO_F401RE | NUCLEO_F401RE[0]     | COM273      | D:          | 072002210774650F3830F312                         |
        | K64F          | K64F[0]              | COM255      | E:          | 0240000028884e450037700f6bf0000d8021000097969900 |
        +---------------+----------------------+-------------+-------------+--------------------------------------------------+
mbedgt: processing target 'K64F' toolchain 'GCC_ARM' compatible platforms... (note: switch set to --parallel 1)
        +---------------+----------------------+-------------+-------------+--------------------------------------------------+
        | platform_name | platform_name_unique | serial_port | mount_point | target_id                                        |
        +---------------+----------------------+-------------+-------------+--------------------------------------------------+
        | K64F          | K64F[0]              | COM255:9600 | E:          | 0240000028884e450037700f6bf0000d8021000097969900 |
        +---------------+----------------------+-------------+-------------+--------------------------------------------------+
mbedgt: running 2 tests for platform 'K64F' and toolchain 'GCC_ARM'
        use 1 instance of execution threads for testing
mbedgt: checking for 'host_tests' directory above image directory structure
        found 'host_tests' directory in: 'TESTS\host_tests'
mbedgt: selecting test case observer...
[...]

Issue #185, opened 2016-09-06
mbedgt  with multiple target: wrong platform_name


jeromecoutant:
At the end of mbedgt command, we got some test result summary

mbedgt: test case report:
+-----------------------+---------------+------------------------------------------------------------------------------+------------------------------------------------------------------------------+--------+--------+--------+--------------------+
| target                | platform_name | test suite                                                                   | test case                                                                    | passed | failed | result | elapsed_time (sec) |
+-----------------------+---------------+------------------------------------------------------------------------------+------------------------------------------------------------------------------+--------+--------+--------+--------------------+
| DISCO_L476VG-ARM      | DISCO_L476VG  | features-frameworks-utest-tests-unit_tests-basic_test                        | Repeating Test                                                               | 2      | 0      | OK     | 0.08               |

First target is OK, but then:

| NUCLEO_L476RG-IAR     | DISCO_L476VG  | features-frameworks-utest-tests-unit_tests-basic_test                        | features-frameworks-utest-tests-unit_tests-basic_test                        | 0      | 1      | ERROR  | 67.71              |

==>  platform_name is not correct… But maybe this issue is only typo?

ciarmcom:
ARM Internal Ref: IOTSFW-2900

PR #189, opened 2016-09-07
Multi-target runs now copy to the correct device

bridadan:
This PR addresses #187. Previously if you specified multiple targets with the `-t` option, it would not copy the binary to the correct device, for example if you ran:

```
mbedgt -V -v -t K64F-GCC_ARM,NUCLEO_F401RE-GCC_ARM
```

The first test run would copy to the K64F correctly, but when the next run was supposed to copy to the NUCLEO_F401RE, it would copy to the K64F instead.

Looks like the issue was the `muts_to_test` variable. This should change between test builds, however it was being kept in global scope. I've now moved it withing the scope of the test build loop so it should reset with each iteration.

### DIFF
--- a/mbed_greentea/mbed_greentea_cli.py
+++ b/mbed_greentea/mbed_greentea_cli.py
@@ -738,7 +738,6 @@ def main_cli(opts, args, gt_instance_uuid=None):
     target_platforms_match = 0  # Count how many platforms were actually tested with current settings
 
     test_report = {}            # Test report used to export to Junit, HTML etc...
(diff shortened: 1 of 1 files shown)


adbridge:
LGTM +1

jeromecoutant:
This should also address https://github.com/ARMmbed/greentea/issues/185 ?

bridadan:
@jeromecoutant It is quite possible that this also fixes that issue. Would you mind updating your issue if you find that the latest master of greentea fixes this?

Issue #186, opened 2016-09-06
Multi-target test run KeyError with differing test cases


bridadan:
If you do a multi-target test run (ex `mbedgt -V -v -t K64F-GCC_ARM,NUCLEO_F401RE-GCC_ARM --report-html=report.html`) and attempt to save an HTML report, you get the following traceback:

```
mbedgt: all tests finished!
mbedgt: shuffle seed: 0.9936295739
mbedgt: exporting to HTML file 'report.html'...
mbedgt: unexpected error:
        u'features-net-feature_ipv4-tests-mbedmicro-net-udp_echo_client'
Traceback (most recent call last):
  File "C:\Python27\Scripts\mbedgt-script.py", line 9, in <module>
    load_entry_point('mbed-greentea', 'console_scripts', 'mbedgt')()
  File "c:\users\bridan01\documents\dev\greentea\mbed_greentea\mbed_greentea_cli.py", line 389, in main
    cli_ret = main_cli(opts, args)
  File "c:\users\bridan01\documents\dev\greentea\mbed_greentea\mbed_greentea_cli.py", line 1007, in main_cli
    html_report = exporter_html(test_report)
  File "c:\users\bridan01\documents\dev\greentea\mbed_greentea\mbed_report_api.py", line 646, in exporter_html
    test_results = test_result_ext["%s-%s" % (platform, toolchain)][test_name]
KeyError: u'features-net-feature_ipv4-tests-mbedmicro-net-udp_echo_client'
```

ciarmcom:
ARM Internal Ref: IOTSFW-2899

bridadan:
Should be fixed with #190

PR #190, opened 2016-09-07
Handle case where tests aren't in each target run

bridadan:
This was first reported by @jeromecoutant, thanks!

This addresses #186. The root cause of the issue was when exporting HTML reports, it was assumed that all the same test cases would be run on all the targets. However, this is not always true. This PR adds a 'NOT_RAN` box to a test that did not run on the current target but DID run on at least one target.

super_awesome_not_ran_html_report

Please review @ConorPKeegan
FYI @mazimkhan


ConorPKeegan:
@bridadan LGTM

Issue #198, opened 2016-10-14
Python error while trying to test NUCLEO_F429ZI and latest mbed-greentea


screamerbg:
### Platform: Windows 7

```
>python --version
Python 2.7.9
>mbedgt --version
1.2.2
>mbed --version
0.9.9
>mbedls --version
1.2.8
```

Also

```
>mbed detect

[mbed] Detected NUCLEO_F429ZI, port COM73, mounted E:
[mbed] Supported toolchains for NUCLEO_F429ZI
+---------------+-----------+-----------+-----------+-----------+-----------+
| Target        | mbed OS 2 | mbed OS 5 |    ARM    |  GCC_ARM  |    IAR    |
+---------------+-----------+-----------+-----------+-----------+-----------+
| NUCLEO_F429ZI | Supported | Supported | Supported | Supported | Supported |
+---------------+-----------+-----------+-----------+-----------+-----------+
Supported targets: 1
Supported toolchains: 3
```

And

```
>mbedls
+---------------+----------------------+-------------+-------------+--------------------------+-----------------+
| platform_name | platform_name_unique | mount_point | serial_port | target_id                | daplink_version |
+---------------+----------------------+-------------+-------------+--------------------------+-----------------+
| NUCLEO_F429ZI | NUCLEO_F429ZI[0]     | E:          | COM73       | 07960221002962073F6DF4AC | 0221            |
+---------------+----------------------+-------------+-------------+--------------------------+-----------------+
```

Running `mbed test -t ARM -m NUCLEO_F429ZI`

```
Build successes:
  * NUCLEO_F429ZI::ARM::FEATURES-FEATURE_LWIP-TESTS-MBEDMICRO-NET-NIST_INTERNET_TIME_SERVICE
  * NUCLEO_F429ZI::ARM::FEATURES-FEATURE_LWIP-TESTS-MBEDMICRO-NET-TCP_CLIENT_ECHO
  * NUCLEO_F429ZI::ARM::FEATURES-FEATURE_LWIP-TESTS-MBEDMICRO-NET-TCP_CLIENT_HELLO_WORLD
  * NUCLEO_F429ZI::ARM::FEATURES-FEATURE_LWIP-TESTS-MBEDMICRO-NET-UDP_ECHO_CLIENT
  * NUCLEO_F429ZI::ARM::FEATURES-FRAMEWORKS-UTEST-TESTS-UNIT_TESTS-BASIC_TEST
[...]

PR #201, opened 2016-11-02
Fixes issue #198 where an error occurs when trying to use the endswith()

adbridge:
method on a NoneType.

On windows 7 if a target was detected but for some reason the serial port
was not, then mbed-ls would return a NoneType for the serial port field in
the ready_mbed_devices list. This serial port field under normal
circumstances will contain a string and has the endswith() method called
on it. This is however illegal for a NoneType.
The fix is to read the serial port field first and test whether or not it
is the NoneType prior to calling the method. If it is the NoneType then
report an error and skip this target.

### DIFF
--- a/mbed_greentea/mbed_greentea_cli.py
+++ b/mbed_greentea/mbed_greentea_cli.py
@@ -778,17 +778,18 @@ def main_cli(opts, args, gt_instance_uuid=None):
             if accepted_target_ids and mbed_dev['target_id'] not in accepted_target_ids:
                 continue
 
+            # Check that we have a valid serial port detected.
+            sp = mbed_dev['serial_port']
+            if not sp:
+                gt_logger.gt_log_err("Serial port for target %s not detected correctly\n" % mbed_dev['target_id'])
+                continue
+
             if mbed_dev['platform_name'] == platform_name:
                 # We will force configuration specific baudrate by adding baudrate to serial port
                 # Only add baudrate decoration for serial port if it's not already there
                 # Format used by mbedhtrun: 'serial_port' = '<serial_port_name>:<baudrate>'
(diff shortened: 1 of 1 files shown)


adbridge:
@mazimkhan , @bridadan please review

mbed_greentea/mbed_greentea_cli.py
```diff
@@ -782,8 +782,14 @@ def get_parallel_value(value):
                 # We will force configuration specific baudrate by adding baudrate to serial port
                 # Only add baudrate decoration for serial port if it's not already there
                 # Format used by mbedhtrun: 'serial_port' = '<serial_port_name>:<baudrate>'
-                if not mbed_dev['serial_port'].endswith(str(baudrate)):
-                    mbed_dev['serial_port'] = "%s:%d" % (mbed_dev['serial_port'], baudrate)
+                sp = mbed_dev['serial_port']
+                if sp:
```
mazimkhan: Instead can we add this check to the check at line https://github.com/ARMmbed/greentea/pull/201/files#diff-dede25ea3355a799ef1ec00a411ffd73R778

Here it gives the impression that checking serial port is related to annotating serial port with baud rate. But then in else we have ```continue```.

adbridge: ahh so just put the check at the top with the accepted_target_ids. Yes can do that.

adbridge:
@mazimkhan please re-review

mazimkhan:
+1 LGTM

adbridge:
Installed locally and running

adbridge:
Local tests ran successfully so merging

adbridge:
Fixes #198

PR #212, opened 2017-02-02
Added pass rates in HTML report

mray19027:
This PR modifies the top level HTML page generated during a test specified with parameters: --report-html=output.html. The intent for the modifications is to create a cleaner report for giving to clients or executives with easy to read pass rates for each test.

Each test now contains a pass rate, along with a ratio of total tests passed to total tests.

CC: @bridadan @BlackstoneEngineering

### DIFF
--- a/mbed_greentea/mbed_report_api.py
+++ b/mbed_greentea/mbed_report_api.py
@@ -354,7 +354,7 @@ html_template = """
         <table>
         <colgroup>
             <col width="auto">
-            <col span="%d" width="150px">
+            <col span="%d" width="200px">
         </colgroup>
             %s
         </table>
@@ -587,7 +587,7 @@ def exporter_html(test_result_ext, test_suite_properties=None):
     result_cell_template = """
                 <td>
                     <div class="result %s" onclick="toggleOverlay('%s')">
-                        <center>%s</center>
+                        <center>%s  -  %s&#37; (%s/%s)</center>
                         %s
                     </div>
                 </td>"""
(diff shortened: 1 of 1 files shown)


bridadan:
Oooooh boy that report sure is purty! :smile: Nice work @mray19027!

greentea_report

One thing I noticed is if a test crashes/fails it doesn't properly reflect the total number of tests.

greentea_report_timeout_pr

That test actually contains 16 test cases, so it should really be something like 25% (4/16)

Looking at the current report, it doesn't list the other tests either, but it also doesn't show the 100%:

greentea_report_timeout_master

So it probably shouldn't block this from coming in, but if there's a quick fix for this I think it would improve the experience quite a bit!

bridadan:
> That test actually contains 16 test cases, so it should really be something like 25% (4/16)

Turns out that Greentea doesn't pass these other 16 test cases to the report APIs, so this will need to be fixed another PR. LGTM! :+1:

BlackstoneEngineering:
LGTM please!

mbed_greentea/mbed_report_api.py
```diff
@@ -672,11 +734,19 @@ def exporter_html(test_result_ext, test_suite_properties=None):
...
+                for index, (testcase_result_name, testcase_result) in enumerate(test_results['testcase_result'].iteritems()):
+                    test_results['single_test_passes'] += testcase_result['passed']
+                    test_results['single_test_count'] += 1
 
                 result_class = get_result_colour_class(test_results['single_test_result'])             
                 this_row += result_cell_template % (result_class,
                                                     result_div_id,
                                                     test_results['single_test_result'],
+                                                    int((test_results['single_test_passes']*100.0)/test_results['single_test_count']),
+                                                    test_results['single_test_passes'],
```
jeromecoutant: Hi
Please, have a look on https://github.com/ARMmbed/greentea/issues/236
This new division is crashing several non regression test sessions on our test bench....
Thx

PR #219, opened 2017-04-05
Set -P to default 300 sec when using grm.

studavekar:
It takes a longer time when running gt with grm option due to network
operations, setting it a higher value with sufficient tolerance.

Todo - Should additional argument be added to GT to pass timeout 
parameter to htrun.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,8 @@ Greentea is the automated testing tool for mbed OS development. It automates the
 
 This document should help you start using Greentea. Please see the [htrun documentation](https://github.com/ARMmbed/htrun), the tool Greentea uses to drive tests, for the technical details of the interactions between the platform and the host machine.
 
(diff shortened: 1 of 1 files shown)


mbed_greentea/mbed_test_api.py
```diff
@@ -244,6 +244,7 @@ def get_binary_host_tests_dir(binary_path, level=2):
         # Example:
         # $ mbedhtrun -p :9600 -f "tests-mbed_drivers-generic_tests.bin" -m K64F --grm raas_client:10.2.203.31:8000
         cmd += ['--grm', global_resource_mgr]
+        cmd += ['-P',"300"] #required higher timeout  due to network operations.
```
jupe: what is default timeout?

studavekar: Default is 60 sec

jupe: is it generally for every tests? sounds a bit short especially for hw tests. is there another mechanism to configure timeout when test takes longer?

PR #223, opened 2017-05-24
Add total stack usage to memory matrix.

studavekar:
This change adds total stack usage which is sum of max stack used by each
thread.

### DIFF
--- a/mbed_greentea/mbed_report_api.py
+++ b/mbed_greentea/mbed_report_api.py
@@ -786,6 +786,10 @@ def exporter_memory_metrics_csv(test_result_ext, test_suite_properties=None):
                         report_key = '%s_%s_max_stack_usage' % (target_name, test_suite_name)
                         metrics_report[report_key] = thread_stack_summary['max_stack_usage']
 
+                    if 'max_stack_usage_total' in thread_stack_summary:
+                        report_key = '%s_%s_max_stack_usage_total' % (target_name, test_suite_name)
+                        metrics_report[report_key] = thread_stack_summary['max_stack_usage_total']
+
     column_names = sorted(metrics_report.keys())
     column_values = [str(metrics_report[x]) for x in column_names]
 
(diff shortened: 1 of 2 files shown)


mbed_greentea/mbed_report_api.py
```diff
@@ -768,6 +770,10 @@ def exporter_memory_metrics_csv(test_result_ext, test_suite_properties=None):
         for test_suite_name in test_results:
             test = test_results[test_suite_name]
 
+            # If test didn't pass, memory stats doesn't make much sense
+            if test['single_test_result'] != TEST_RESULT_OK:
+                continue
```
bridadan: @studavekar Seems like this is pretty harmless to leave in anyway right? Maybe the test failed due to a stack overflow...

studavekar: Yeah true that is good candidate where if test fails we would want the results of stack. However for failing tests the memory data can't be truly trusted. If test fails at init_xxx() the data reported by it would smaller. So for reporting we are ignoring the fail tests.

Can take out this check so we get data in either case.

mbed_greentea/mbed_report_api.py
```diff
@@ -17,6 +17,8 @@
 Author: Przemyslaw Wirkus <Przemyslaw.wirkus@arm.com>
 """
 
+from mbed_test_api import TEST_RESULT_OK
```
bridadan: One last nit pick! This line isn't necessary any more since you removed that last check. Mind pulling this out too?

studavekar: done

mbed_greentea/mbed_test_api.py
```diff
@@ -551,20 +551,22 @@ def get_memory_metrics(output):
...
     max_thread_stack_size = 0
     max_thread = None
-
+    max_stack_usage_total = 0
     for cur_thread_stack_info in thread_stack_info:
         if cur_thread_stack_info['stack_size'] > max_thread_stack_size:
             max_thread_stack_size = cur_thread_stack_info['stack_size']
             max_thread = cur_thread_stack_info
-
+        max_stack_usage_total += cur_thread_stack_info['max_stack']
```
studavekar: updated it to correctly report max_stage_usage_total

PR #224, opened 2017-05-26
SYNC failure case added: Dependent on HTRUN code

deepikabhavnani:
@bridadan @mazimkhan @studavekar : Please review code

Many boards in CI fail just after sync, suggestion to which was re-flash firmware and re-try. Capturing SYNC failure as separate for such cases.
Also would like to know more about SYNC=2 value, increasing that will have impact on anything?\

Note: Code changes in htrun should be pulled along with this request


bridadan:
> Also would like to know more about SYNC=2 value, increasing that will have impact on anything?

The impact of that is the number of `__sync` messages that are sent to the device. So you could keep increasing that, but the number really should be fairly low if the device is behaving reasonably. Most devices shouldn't take multiple seconds to boot.

bridadan:
Note, when this gets merged, the release should bump its dependency on `mbed-host-tests` to the next release version. It's currently on 1.1.8, so I'm guessing it will be 1.1.9.

PR #225, opened 2017-05-31
Getting copy and reset methods from mbed-os targets.json.

bridadan:
This commit allows targets to specify custom "copy" and "reset" methods
when testing with Greentea. This method should not be used extensively.
It is intended to be used by targets that don't have a typical mbed
interface.

It is needed to enable targets like the ARM MPS2: https://developer.mbed.org/platforms/ARM-MPS2/

FYI @ashok-rao @screamerbg 

Could you please take a look @adbridge @mazimkhan


mazimkhan:
Its really useful. Thanks for this.

Issue #150, opened 2016-06-23
Ctrl+C when running tests doesn't stop the tests (Linux)


MarceloSalazar:
- Type: Bug
- Priority:  Major

Raised earlier in wrong repo: https://github.com/ARMmbed/mbed-os/issues/387
## Bug

Under Linux Ubuntu 14.04, the command ctrl+C doesn't (fully) stop the tests. To reproduce the problem, just run `mbed test` under linux and try to stop it by using ctrl+C.

I get access to the console, but tests continue to run.

The processes continue to run in the background:

```
19651 pts/12   00:00:01 mbedgt
21283 pts/12   00:00:02 mbedhtrun
21299 pts/12   00:00:00 mbedhtrun
```

The workaround is to manually kill the processes.

PrzemekWirkus:
@mlnx Please let me know following:
```
$ mbed-ls --version
$ mbed-host-tests --version
$ mbed-greentea --version
$ python --version
```

ciarmcom:
ARM Internal Ref: IOTSFW-2744

mlnx:
```
$ mbedls --version
0.2.13
$ mbedhtrun --version
0.2.18
$ mbedgt --version
0.2.24
$ python --version
Python 2.7.6
```

bridadan:
This should be fixed in #226, feel free to try out the fix if you like.

theotherjimmy:
Closing as it's fixed.

PR #226, opened 2017-06-06
Fixing keyboard interrupts on posix OSes

bridadan:
This should fix issue #150. Tested working on Windows 10 and Ubuntu Linux 16.04.

The underlying issue was that a blocking call to `join` for the worker threads was stopping `KeyboardInterrupt` exceptions from being raised to the exception handler here: https://github.com/ARMmbed/greentea/blob/master/mbed_greentea/mbed_greentea_cli.py#L396

This fix changes the blocking call to `join` to one that timesout after 0.1 seconds. The join is called in a loop until the thread's `isAlive()` function returns False.

@geky @theotherjimmy @jupe I'm sure you'll be very happy to see this fix finally happen 😄 

@mazimkhan Could you please take a look?

### DIFF
--- a/mbed_greentea/mbed_greentea_cli.py
+++ b/mbed_greentea/mbed_greentea_cli.py
@@ -906,7 +906,15 @@ def main_cli(opts, args, gt_instance_uuid=None):
         # merge partial test reports from different threads to final test report
         for t in execute_threads:
             try:
-                t.join() #blocking
+                # We can't block forever here since that prevents KeyboardInterrupts
+                # from being propagated correctly. Therefore, we just join with a
+                # timeout of 0.1 seconds until the thread isn't alive anymore.
+                # A time of 0.1 seconds is a fairly arbitrary choice. It needs
+                # to balance CPU utilization and responsiveness to keyboard interrupts.
+                # Checking 10 times a second seems to be stable and responsive.
+                while t.isAlive():
+                    t.join(0.1)
+
                 test_return_data = test_result_queue.get(False)
             except Exception as e:
                 # No test report generated


mbed_greentea/mbed_greentea_cli.py
```diff
@@ -906,7 +906,12 @@ def get_parallel_value(value):
         # merge partial test reports from different threads to final test report
         for t in execute_threads:
             try:
-                t.join() #blocking
+                # We can't block forever here since that prevents KeyboardInterrupts
+                # from being propagated correctly. Therefore, we just join with a
+                # timeout of 0.1 seconds until the thread isn't alive anymore.
+                while t.isAlive():
+                    t.join(0.1)
```
adbridge: Any particular reason for choosing 0.1 secs ?

mazimkhan:
Looks good to me. Though isn't 100ms too aggressive for an unlikely case.

bridadan:
The `0.1` seconds was arbitrary I admit. It just needs to balance CPU utilization and responsiveness to the users Ctrl+C keyboard stroke. This should be a very manageable load on the CPU (on Windows it was using <1% during testing). It should also still feel very responsive (vs a 1 second poll period for instance).

> Though isn't 100ms too aggressive for an unlikely case.

@mazimkhan Not sure what the aggressive case is, can you explain?

adbridge:
@bridadan Might be worth adding something along those lines to the code

bridadan:
@adbridge comments amended!

adbridge:
@mazimkhan are you happy with this?

mazimkhan:
Yup looks goods.

adbridge:
@bridadan did you test this worked on other OS' ? I tried in in the mbed2 release script running on gitbash (windows) and it failed unless I also had 
        thr.setDaemon(True)
for each thread...

bridadan:
@adbridge I tested on Windows 10 and Ubuntu Linux 16.04. What OS are you testing on? It may be that we have to add that as well. Please let me know how to reproduce!

adbridge:
@bridadan Shell = gitbash, running on windows 10 . I had multiple concurrent threads in the mbed2 release script so I don't know if that makes a difference either.

bridadan:
@adbridge Sorry I'm still not 100% sure what you mean. Are you saying that when you were implementing multithreading in the release script that you had to use `thread.setDaemon(True)` for it to work? If so, it looks like we're already setting these threads to be daemon threads: https://github.com/bridadan/greentea/blob/bf221e28ed1d9a00f0179140e854309d7ad9a018/mbed_greentea/mbed_greentea_cli.py#L903

Or did you try this PR locally and it still didn't work?

adbridge:
@bridadan the former :) . I only looked at the actual patch not what was there previously , so if the threads were already set up as daemons then you're probably fine! :)

Issue #231, opened 2017-07-10
Tests reported as skipped when they fail due to timeout


k-stachowiak:
When launching an mbedtls test suite via htrun a bug in one of the communication scripts leading the test data being sent incorrectly (the said bug is irrelevant to this report). This lead to the script fail on target, but from the host's point of view a timeout was observed. This was properly reported in the full log, but the test was reported as SKIPPED in the summary report.

In such case the system should report FAIL instead of SKIPPED.

bridadan:
Hi @k-stachowiak, can you post the full verbose (add `-v` to `mbed test` or `-V` if executing directly with `mbedgt`) test log here? Depending on how the test failed, it may report the other testcases as SKIPPED. However I would agree that if the test case that timedout should report FAIL/TIMEOUT.

k-stachowiak:
I included the most relevant log pieces. If it turns out not to be enough please let me know, I will strip it of any sensitive information it may contain and provide it. Below are the most of the details that I can give you quickly.

The behavior is triggered by intentionally corrupting the binary data transmission channel by mismatching the byte alignment. We investigated two cases: 4byte alignment on the host, 2bytes on the target, and 2bytes on the host, 4bytes on the target. I believe that the way we trigger the behavior, even though rather pathological and not likely in normal execution, is irrelevant. The issue isn't critical, but seems worth an investigation. Let me provide more details.

I am running the following command to run the tests:

    mbedgt -e . --test-spec testspec.json -g {device type}:module_name:10.2.123.43:3334 -v -V -c {device type} -p COM6

* When the host is 4 bytes and target 2 we observe the DISPATCH_INVALID_TEST_DATA error for several tests, which then are reported as skipped in the summary. For example:

```
[1499760004.70][HTST][INF] Running: Test bit set (Change existing value with a 1)
[1499760004.71][SERI][TXD] {{   H; I
S       49979687 I     I      I
S       66756903 I      }}
[1499760004.73][CONN][INF] found KV pair in stream: {{F;-4}}, queued...
[1499760004.74][HTST][INF] Error: DISPATCH_INVALID_TEST_DATA
```

And in the report:

    test case: 'Test bit set (Change existing value with a 1)' ................................... SKIPPED in 0.00 sec

[...]

PR #232, opened 2017-07-11
Fixing regex that caused testcase names to be missed.

bridadan:
Handled key value pairs stopped being printed in their raw received form
in https://github.com/ARMmbed/htrun/pull/130. This broke the testcase name detection in greentea.
This changes the regex to detect the testcase names in both old and new
versions of htrun.

This should address #231 I believe.

cc @mazimkhan

### DIFF
--- a/mbed_greentea/mbed_test_api.py
+++ b/mbed_greentea/mbed_test_api.py
@@ -341,8 +341,8 @@ def get_testcase_count_and_names(output):
     testcase_count = 0
     testcase_names = []
 
-    re_tc_count = re.compile(r"^\[(\d+\.\d+)\]\[(\w+)\]\[(\w+)\] \{\{(__testcase_count);(\d+)\}\}")
-    re_tc_names = re.compile(r"^\[(\d+\.\d+)\]\[(\w+)\]\[(\w+)\] \{\{(__testcase_name);([^;]+)\}\}")
+    re_tc_count = re.compile(r"^\[(\d+\.\d+)\]\[(\w+)\]\[(\w+)\].*\{\{(__testcase_count);(\d+)\}\}")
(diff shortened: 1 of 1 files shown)


mbed_greentea/mbed_test_api.py
```diff
@@ -341,8 +341,8 @@ def get_testcase_count_and_names(output):
     testcase_count = 0
     testcase_names = []
 
-    re_tc_count = re.compile(r"^\[(\d+\.\d+)\]\[(\w+)\]\[(\w+)\] \{\{(__testcase_count);(\d+)\}\}")
-    re_tc_names = re.compile(r"^\[(\d+\.\d+)\]\[(\w+)\]\[(\w+)\] \{\{(__testcase_name);([^;]+)\}\}")
+    re_tc_count = re.compile(r"^\[(\d+\.\d+)\]\[(\w+)\]\[(\w+)\].*\{\{(__testcase_count);(\d+)\}\}")
```
mazimkhan: @bridadan can you please post the log that matches or not matches with existing code. Just to see what the change in this regex helps.

bridadan:
@mazimkhan Sure. Greentea is currently failing to pick up these lines:

```
[1499788617.72][CONN][INF] found KV pair in stream: {{__testcase_name;C strings: strtok}}, queued...
[1499788617.76][CONN][INF] found KV pair in stream: {{__testcase_name;C strings: strpbrk}}, queued...
```

They used to be printed like this:

```
[1499788676.98][CONN][RXD] {{__testcase_name;C strings: strtok}}
[1499788676.98][CONN][INF] found KV pair in stream: {{__testcase_name;C strings: strtok}}, queued...
[1499788677.02][CONN][RXD] {{__testcase_name;C strings: strpbrk}}
[1499788677.02][CONN][INF] found KV pair in stream: {{__testcase_name;C strings: strpbrk}}, queued...
```

So this change allows it to pick up both versions

mazimkhan:
LGTM

PR #233, opened 2017-08-04
take sync flag arguments and pass to mbedhtrun

studavekar:
Due to transport overhead we may need to send more then 2 sync packets,
as of now using greentea doesn't have this option. This patch will enable us to
send --sync from greentea.

### DIFF
--- a/mbed_greentea/mbed_greentea_cli.py
+++ b/mbed_greentea/mbed_greentea_cli.py
@@ -281,6 +281,11 @@ def main():
                     default=None,
                     help='Shuffle seed (If you want to reproduce your shuffle order please use seed provided in test summary)')
 
+    parser.add_option('', '--sync',
+                    dest='sync_packet',
+                    default=5,
+                    help='Define how many times __sync packet will be sent to device: 0: none; -1: forever; 1,2,3... - number of  times (Default 2 time)')
+
     parser.add_option('', '--lock',
                     dest='lock_by_target',
                     default=False,
@@ -465,6 +470,7 @@ def run_test_thread(test_result_queue, test_queue, opts, mut, build, build_path,
                                          json_test_cfg=opts.json_test_configuration,
                                          enum_host_tests_path=enum_host_tests_path,
                                          global_resource_mgr=opts.global_resource_mgr,
+                                         sync_packet=opts.sync_packet,
                                          verbose=verbose)
 
         # Some error in htrun, abort test execution
--- a/mbed_greentea/mbed_greentea_cli.py
+++ b/mbed_greentea/mbed_greentea_cli.py
@@ -282,9 +282,9 @@ def main():
                     help='Shuffle seed (If you want to reproduce your shuffle order please use seed provided in test summary)')
 
     parser.add_option('', '--sync',
-                    dest='sync_packet',
+                    dest='num_sync_packtes',
                     default=5,
-                    help='Define how many times __sync packet will be sent to device: 0: none; -1: forever; 1,2,3... - number of  times (Default 5 time)')
+                    help='Define how many times __sync packet will be sent to device: 0: none; -1: forever; 1,2,3... - number of  times (the default is 5 packets)')
 
     parser.add_option('', '--lock',
                     dest='lock_by_target',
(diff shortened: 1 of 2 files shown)


mbed_greentea/mbed_greentea_cli.py
```diff
@@ -281,6 +281,11 @@ def main():
                     default=None,
                     help='Shuffle seed (If you want to reproduce your shuffle order please use seed provided in test summary)')
 
+    parser.add_option('', '--sync',
+                    dest='sync_packet',
+                    default=5,
```
bridadan: Would you like to increase this value to 5? Or should we keep it at 2 for now? If you'd like to change it to 5, please update the help text below.

studavekar:
@bridadan updated the help.

jupe:
How long time it waits before sending another sync packets? Is it documented? Could there be another solution to synchronous DUT like waiting some message from DUT so that user doesn't need to take care of this kind of thing? just thoughts..

bridadan:
@jupe I don't think its documented at the moment, [but the delay is 1 second](https://github.com/ARMmbed/htrun/blob/master/mbed_host_tests/host_tests_conn_proxy/conn_proxy.py#L136).

There could probably be another solution, but lets keep this PR on topic. Feel free to open up an issue if you want to discuss alternatives. Generally a user doesn't have to take care of this kind of thing, it usually just works. I don't think I've actually ever seen a case where the sync was sent "too fast".

studavekar:
@mazimkhan can you please review

mbed_greentea/mbed_greentea_cli.py
```diff
@@ -281,6 +281,11 @@ def main():
                     default=None,
                     help='Shuffle seed (If you want to reproduce your shuffle order please use seed provided in test summary)')
 
+    parser.add_option('', '--sync',
+                    dest='sync_packet',
```
theotherjimmy: This is not a good name for this parameter. `sync_num` or `num_sync_packtes` or something like that is a better name.

mbed_greentea/mbed_greentea_cli.py
```diff
@@ -281,6 +281,11 @@ def main():
...
+                    default=5,
+                    help='Define how many times __sync packet will be sent to device: 0: none; -1: forever; 1,2,3... - number of  times (Default 5 time)')
```
theotherjimmy: Nit: (**the** default is 5 **packets**)

studavekar:
@theotherjimmy have incorporated review changes.

--sync is the argument to htrun, have kept the same for consistency 

```

(env) arm@austin-ci-linux-002:~/repo/greentea$ mbedhtrun --help
Usage: mbedhtrun [options]

Flash, reset and perform host supervised tests on mbed platforms

Options:
  -h, --help            show this help message and exit
  -m MICRO, --micro=MICRO
                        Target microcontroller name
  -p PORT, --port=PORT  Serial port of the target
  -d DISK_PATH, --disk=DISK_PATH
                        Target disk (mount point) path
  -t TARGET_ID, --target-id=TARGET_ID
                        Unique Target Id or mbed platform
  --sync=SYNC_BEHAVIOR  Define how many times __sync packet will be sent to
                        device: 0: none; -1: forever; 1,2,3... - number of
                        times (Default 2 time)

```

mbed_greentea/mbed_greentea_cli.py
```diff
@@ -281,6 +281,11 @@ def main():
                     default=None,
                     help='Shuffle seed (If you want to reproduce your shuffle order please use seed provided in test summary)')
 
+    parser.add_option('', '--sync',
+                    dest='num_sync_packtes',
+                    default=5,
+                    help='Define how many times __sync packet will be sent to device: 0: none; -1: forever; 1,2,3... - number of  times (the default is 5 packet)')
```
theotherjimmy: You dropped the "s" from "packets"

theotherjimmy:
@studavekar Thanks for explaining your naming convention. Of course consistency is more important here.

mbed_greentea/mbed_greentea_cli.py
studavekar: updated "s"  :)

theotherjimmy:
@studavekar do you have any reservations with merging this now?

studavekar:
Lets merge it 👍

PR #234, opened 2017-08-07
Add option to request allocation based on specified tags

kegilbert:
In the process of integrating additional tests to the CI we realized it would be helpful to be able to tag particular devices in RaaS that had unique or additional setups for tests that required them (for example a K64F with an ESP8266 or Ethernet). 

## Usage 
Take the following tag as an example. 

image

There is a [separate htrun PR](https://github.com/ARMmbed/htrun/pull/163) up to implement this change, this PR will funnel the new argument to the htrun layer. Given that htrun change and the above tag setup, we could run tests with the following setup.

## Test Setup
```
mbedgt -g "K64F:raas_client:austin-ci-linux-001:8000" --test-spec=BUILD/tests/K64F/GCC_ARM/test_spec.json --report-html=K64F_GCC_ARM.html -e tests/host_tests/timing_drift_auto.py -v -V -n features-feature_lwip-tests-mbedmicro-net-tcp_hello_world --allocate-on-tag Ethernet                                                             
 ```

## Multiple Tags
Comma separated lists can also be supplied if multiple tags are required (device resource must match *all* of the supplied tags, it cannot be used as an 'or' list).
```
.... --allocate-on-tag Ethernet,ESP8266,...
```

### DIFF
--- a/mbed_greentea/mbed_greentea_cli.py
+++ b/mbed_greentea/mbed_greentea_cli.py
@@ -286,6 +286,11 @@ def main():
                     default=5,
                     help='Define how many times __sync packet will be sent to device: 0: none; -1: forever; 1,2,3... - number of  times (the default is 5 packets)')
 
+    parser.add_option('', '--tag-filters',
+                    dest='tags',
+                    default=None,
+                    help='Filter list of available devices under test to only run on devices with the provided list of tags  [tag-filters tag1,tag]')
+
     parser.add_option('', '--lock',
                     dest='lock_by_target',
                     default=False,
@@ -471,6 +476,7 @@ def run_test_thread(test_result_queue, test_queue, opts, mut, build, build_path,
                                          enum_host_tests_path=enum_host_tests_path,
                                          global_resource_mgr=opts.global_resource_mgr,
                                          num_sync_packtes=opts.num_sync_packtes,
+                                         tags=opts.tags,
                                          verbose=verbose)
 
         # Some error in htrun, abort test execution
--- a/mbed_greentea/mbed_test_api.py
+++ b/mbed_greentea/mbed_test_api.py
@@ -122,6 +122,7 @@ def run_host_test(image_path,
                   enum_host_tests_path=None,
                   global_resource_mgr=None,
                   num_sync_packtes=None,
+                  tags=None,
                   run_app=None):
     """! This function runs host test supervisor (executes mbedhtrun) and checks output from host test process.
     @param image_path Path to binary file for flashing
@@ -138,6 +139,8 @@ def run_host_test(image_path,
     @param max_failed_properties After how many unknown properties we will assume test is not ported
     @param enum_host_tests_path Directory where locally defined host tests may reside
     @param num_sync_packtes sync packets to send for host <---> device communication
+    @param tags Filter list of available devices under test to only run on devices with the provided list
+           of tags  [tag-filters tag1,tag]
     @param run_app Run application mode flag (we run application and grab serial port data)
     @param digest_source if None mbedhtrun will be executed. If 'stdin',
            stdin will be used via StdInObserver or file (if
@@ -273,6 +276,8 @@ def run_host_test(image_path,
 
     if num_sync_packtes:
         cmd += ["--sync",str(num_sync_packtes)]
+    if tags:
+        cmd += ["--tag-filters", tags]
 
     gt_logger.gt_log_tab("calling mbedhtrun: %s"% " ".join(cmd), print_text=verbose)
     gt_logger.gt_log("mbed-host-test-runner: started")

### CODE REVIEW

**Comment by kegilbert** [comment]
cc @studavekar

File: mbed_greentea/mbed_test_api.py [hunk 1]
```diff
@@ -269,6 +272,9 @@ def get_binary_host_tests_dir(binary_path, level=2):
         if run_app:
             cmd += ["--run"]    # -f stores binary name!
 
+    if tags:
+        cmd+= ["--allocate-on-tag", str(tags)]
+
```
studavekar: should this be only passed only if global_resource_mgr is true

kegilbert:
@studavekar updated

theotherjimmy:
It looks like this PR might depend on https://github.com/ARMmbed/htrun/pull/163

mbed_greentea/mbed_greentea_cli.py
```diff
@@ -281,6 +281,11 @@ def main():
                     default=None,
                     help='Shuffle seed (If you want to reproduce your shuffle order please use seed provided in test summary)')
 
+    parser.add_option('', '--allocate-on-tag',
```
theotherjimmy: I'm not convinced that `allocate-on-tag` captures what this does exactly. It seems to be a filter, using a tag, for devices under test. If this is correct, could you change the help text to match this? Could you also change the parameter to better describe it's function? something like `tag-filters` might be a better name.

mbed_greentea/mbed_test_api.py
```diff
@@ -136,6 +137,8 @@ def run_host_test(image_path,
     @param json_test_cfg Additional test configuration file path passed to host tests in JSON format
     @param max_failed_properties After how many unknown properties we will assume test is not ported
     @param enum_host_tests_path Directory where locally defined host tests may reside
+    @param allocate_on_tag Allocate targets only with specified comma seperated list of tags
```
theotherjimmy: `@param` does not match parameter name. Please fix this.

mbed_greentea/mbed_test_api.py
```diff
@@ -247,6 +250,9 @@ def get_binary_host_tests_dir(binary_path, level=2):
         # Example:
         # $ mbedhtrun -p :9600 -f "tests-mbed_drivers-generic_tests.bin" -m K64F --grm raas_client:10.2.203.31:8000
         cmd += ['--grm', global_resource_mgr]
+
+        if tags:
+            cmd+= ["--allocate-on-tag", str(tags)]
```
theotherjimmy: Please add a space in front of the `+=` operator.

theotherjimmy: Further, does `str(tags)` actually correctly format the tags here?

kegilbert:
@theotherjimmy You are correct, this PR is linked to ARMmbed/htrun#163.
Updated this review, working on the htrun PR.

mbed_greentea/mbed_test_api.py
kegilbert: Is there another format that would work better? I haven't had issues parsing the string in the htrun layer once it's passed on here with up to 3 tags specified, but there could be a better implementation.

mbed_greentea/mbed_test_api.py
```diff
@@ -269,6 +272,9 @@ def get_binary_host_tests_dir(binary_path, level=2):
         if run_app:
             cmd += ["--run"]    # -f stores binary name!
 
+    if tags:
+        cmd += ["--tag-filters", str(tags)]
```
theotherjimmy: looking at the code in more depth, it looks like you don't even need the `str` wrapper, as you are not parsing this argument in this command line app.

e.g.
```python
if tags:
    cmd += ["--tag-filters", tags]
```
should work correctly.

theotherjimmy:
@kegilbert I'll make sure to merge this after armmbed/htrun#163.

kegilbert:
Thanks!

theotherjimmy:
@kegilbert It looks like you have merge conflicts. Could you resolve them?

theotherjimmy:
I just merged armmbed/htrun#163, so this should no longer be blocked on anything except the merge conflicts.

kegilbert:
@theotherjimmy rebased and squashed

mbed_greentea/mbed_test_api.py
```diff
@@ -138,6 +139,8 @@ def run_host_test(image_path,
     @param max_failed_properties After how many unknown properties we will assume test is not ported
     @param enum_host_tests_path Directory where locally defined host tests may reside
     @param num_sync_packtes sync packets to send for host <---> device communication
+    @param tag-filters Filter list of available devices under test to only run on devices with the provided list
```
theotherjimmy: This `@param` does not describe a parameter. Perhaps you meant `tags` instead of `tag-filters`?

mbed_greentea/mbed_test_api.py
```diff
@@ -273,6 +276,8 @@ def get_binary_host_tests_dir(binary_path, level=2):
 
     if num_sync_packtes:
         cmd += ["--sync",str(num_sync_packtes)]
+    if tags:
+        cmd += ["--tag-filters", str(tags)]
```
theotherjimmy: Since `tags` here comes from the command line untouched, Could you remove the `str(` and `)` to indicate that it's already in the correct format?

kegilbert:
@theotherjimmy Updated and squashed the branch.

PR #239, opened 2017-09-28
Increase unit test coverage and fix some minor issues

scartmell-arm:
Changes:

- Fix typo in cmake_handlers log output
- Fix typo in greentea_kettle_info() docstring
- Remove unused parameter from get_json_data_from_file()
- Reduce nesting in mbed_target_info to make it easier to test

### DIFF
--- a/mbed_greentea/mbed_report_api.py
+++ b/mbed_greentea/mbed_report_api.py
@@ -157,6 +157,7 @@ def exporter_testcase_junit(test_result_ext, test_suite_properties=None):
     @return String containing Junit XML formatted test result output
     """
     from junit_xml import TestSuite, TestCase
+    import sys
 
     test_suites = []
 
@@ -165,11 +166,11 @@ def exporter_testcase_junit(test_result_ext, test_suite_properties=None):
         for test_suite_name in test_results:
             test = test_results[test_suite_name]
 
-            # tc_elapsed_sec = test['elapsed_time']
-            tc_stdout = str() #test['single_test_output']
-
             try:
-                tc_stdout = test['single_test_output'].decode('unicode_escape').encode('ascii','ignore')
+                if sys.version_info >= (3, 0):
+                    tc_stdout = str(test['single_test_output']).encode('ascii', 'ignore')
+                else:
+                    tc_stdout = test['single_test_output'].decode('unicode_escape').encode('ascii', 'ignore')
             except UnicodeDecodeError as e:
                 err_mgs = "(UnicodeDecodeError) exporter_testcase_junit:", str(e)
                 tc_stdout = err_mgs
@@ -193,7 +194,12 @@ def exporter_testcase_junit(test_result_ext, test_suite_properties=None):
                 result_text = testcase_result[tc_name].get('result_text', "UNDEF")
 
                 try:
-                    tc_stderr = '\n'.join(utest_log).decode('unicode_escape').encode('ascii','ignore')
+                    lines = '\n'.join(utest_log)
+
+                    if sys.version_info >= (3, 0):
+                        tc_stderr = str(lines).encode('ascii', 'ignore')
+                    else:
+                        tc_stderr = lines.decode('unicode_escape').encode('ascii', 'ignore')
                 except UnicodeDecodeError as e:
                     err_mgs = "(UnicodeDecodeError) exporter_testcase_junit:" + str(e)
                     tc_stderr = err_mgs
@@ -218,10 +224,12 @@ def exporter_testcase_junit(test_result_ext, test_suite_properties=None):
                 test_cases.append(tc)
 
             ts_name = target_name
-            if test_suite_properties and target_name in test_suite_properties:
-                test_build_properties = test_suite_properties[target_name] 
+
+            if test_suite_properties is not None:
+                test_build_properties = test_suite_properties[target_name] if target_name in test_suite_properties else None
             else:
                 test_build_properties = None
+
             ts = TestSuite(ts_name, test_cases, properties=test_build_properties)
             test_suites.append(ts)
 
--- a/mbed_greentea/mbed_target_info.py
+++ b/mbed_greentea/mbed_target_info.py
@@ -142,17 +142,10 @@ def parse_yotta_json_for_build_name(yotta_json_content):
     @param yotta_json_content Content of .yotta_json file
     @return String with set yotta target name, None if no target found
     """
-    result = None
-    if 'build' in yotta_json_content:
-        if 'target' in yotta_json_content['build']:
-            # yotta_json_content['build']['target'] should be something like this: "frdm-k64f-gcc,*"
-            # We will split this string using ','. Because split will always return an array
-            # with at least 1 element we can select first element to be target name
-            try:
-                result = yotta_json_content['build']['target'].split(',')[0]
-            except IndexError:
-                result = None
-    return result
+    try:
+        return yotta_json_content['build']['target'].split(',')[0]
+    except KeyError:
+        return None
 
 def get_yotta_target_from_local_config(yotta_json='.yotta.json'):
     """! Load yotta target from local configuration file
@@ -167,15 +160,17 @@ def get_yotta_target_from_local_config(yotta_json='.yotta.json'):
       }
     }
     """
-    result = None
-    if os.path.exists(yotta_json):
-        try:
-            gt_logger.gt_log("parsing local file '%s' for target information"% yotta_json)
-            with open(yotta_json, 'r') as f:
-                result = parse_yotta_json_for_build_name(json.load(f))
-        except Exception as e:
-            print(str(e))
-    return result
+    if not os.path.exists(yotta_json):
+        return None
+
+    try:
+        gt_logger.gt_log("parsing local file '%s' for target information"% yotta_json)
+
+        with open(yotta_json, 'r') as f:
+            return parse_yotta_json_for_build_name(json.load(f))
+    except (IOError, ValueError) as e:
+        gt_logger.gt_log(str(e))
+    return None
 
 def get_mbed_target_from_current_dir():
     """! Function uses yotta target command to check current target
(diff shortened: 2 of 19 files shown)


mbed_greentea/mbed_target_info.py
```diff
@@ -142,17 +142,13 @@ def parse_yotta_json_for_build_name(yotta_json_content):
...
-            # with at least 1 element we can select first element to be target name
-            try:
-                result = yotta_json_content['build']['target'].split(',')[0]
-            except IndexError:
-                result = None
-    return result
+    if 'build' not in yotta_json_content:
+        return None
+
+    if 'target' not in yotta_json_content['build']:
```
theotherjimmy: Can this if be merged with the prior one?

mbed_greentea/mbed_target_info.py
```diff
@@ -142,17 +142,13 @@ def parse_yotta_json_for_build_name(yotta_json_content):
...
+        return None
+
+    return yotta_json_content['build']['target'].split(',')[0]
```
theotherjimmy: I would rewrite this function as:
```python
try:
    return yotta_json_content['build']['target'].split(',')[0]
except KeyError:
    return None
```

mbed_greentea/mbed_target_info.py
```diff
@@ -167,15 +163,17 @@ def get_yotta_target_from_local_config(yotta_json='.yotta.json'):
...
-    return result
+    if not os.path.exists(yotta_json):
+        return None
+
+    try:
+        gt_logger.gt_log("parsing local file '%s' for target information"% yotta_json)
+
+        with open(yotta_json, 'r') as f:
+            return parse_yotta_json_for_build_name(json.load(f))
+    except Exception as e:
```
theotherjimmy: Please refine this except to catch only the exceptions that you intend to generate. As it stands this will eat Control-c and other spurious exceptions.

mbed_greentea/mbed_target_info.py
```diff
@@ -167,15 +163,17 @@ def get_yotta_target_from_local_config(yotta_json='.yotta.json'):
...
+        print(str(e))
```
theotherjimmy: Should this use `gt_logger`?

mbed_greentea/mbed_report_api.py
```diff
@@ -165,11 +166,11 @@ def exporter_testcase_junit(test_result_ext, test_suite_properties=None):
...
 
-            # tc_elapsed_sec = test['elapsed_time']
-            tc_stdout = str() #test['single_test_output']
-
             try:
-                tc_stdout = test['single_test_output'].decode('unicode_escape').encode('ascii','ignore')
+                if sys.version_info >= (3, 0):
+                    tc_stdout = str(test['single_test_output']).encode('ascii', 'ignore')
+                else:
+                    tc_stdout = test['single_test_output'].decode('unicode_escape').encode('ascii', 'ignore')
```
bridadan: Was the previous implementation not working in Python 3? Or does this provide better output?

mbed_greentea/mbed_target_info.py
bridadan: The try/catch block was removed here, was this intentional?

mbed_greentea/mbed_target_info.py
```diff
@@ -223,39 +221,50 @@ def get_mbed_targets_from_yotta_local_module(mbed_classic_name, yotta_targets_pa
...
+        return None
+
+    if not 'keywords' in target_json_data:
+        return None
+
+    for keyword in target_json_data['keywords']:
+        if not keyword.startswith('mbed-target:'):
+            continue
+
+        mbed_target, mbed_name = keyword.split(':')
```
theotherjimmy: I think this may throw an exception if `keyword.split(':')` does not have 2 elements. Should we be skipping these incorrectly formatted entries?

mbed_greentea/mbed_target_info.py
```diff
@@ -304,8 +312,9 @@ def add_target_info_mapping(mbed_classic_name, map_platform_to_yt_target=None, u
         yotta_target_search = list(set(yotta_target_search))    # Reduce repeated values
 
     # Add extra targets to already existing and detected in the system platforms
-    if mbed_classic_name in map_platform_to_yt_target:
-        yotta_target_search = list(set(yotta_target_search + map_platform_to_yt_target[mbed_classic_name]))
+    if map_platform_to_yt_target is not None:
+        if mbed_classic_name in map_platform_to_yt_target:
```
theotherjimmy: Please rewrite this to use one if statement. For example:
```python
if map_paltform_to_yt_target and mbed_classic_name in map_platform_to_yt_target:
    ...
```

mbed_greentea/mbed_target_info.py
```diff
@@ -320,26 +329,32 @@ def add_target_info_mapping(mbed_classic_name, map_platform_to_yt_target=None, u
...
+    mbeds_yt_targets = []
+    for target in target_desc['yotta_targets']:
+        mbeds_yt_targets.append(target['yotta_target'])
+
+    # Check if any of yotta targets is new to TARGET_INFO_MAPPING
+    for new_yt_target in yotta_target_search:
+        if new_yt_target in mbeds_yt_targets:
+            continue
+
+        # gt_logger.gt_log_tab("discovered extra target '%s'"% new_yt_target)
```
theotherjimmy: Could you uncomment or delete this commented code?

mbed_greentea/mbed_test_api.py
```diff
@@ -123,7 +123,7 @@ def run_host_test(image_path,
                   global_resource_mgr=None,
                   num_sync_packtes=None,
                   tags=None,
-                  run_app=None):
+                  run_app=None): # pragma: no cover
```
theotherjimmy: Could you explain why we don't want this function to count towards our coverage?

test/mbed_gt_cmake_handlers.py
```diff
@@ -16,8 +16,17 @@
...
 """
 
+import os
+import mock
+import sys
 import unittest
-from mbed_greentea import cmake_handlers
+
+try:
+    from StringIO import StringIO
```
bridadan: `StringIO` is also available in Python 2's `io` package, so I think you can do away with this try/except block

https://docs.python.org/2/library/io.html#io.StringIO

test/mbed_gt_cli.py
```diff
@@ -38,14 +45,50 @@ def test_get_greentea_version(self):
...
+        test1_dir = tempfile.mkdtemp()
+        test2_dir = os.mkdir(os.path.join(test1_dir, "test"))
+        test3_dir = os.mkdir(os.path.join(test1_dir, "test", "host_tests"))
+
+        os.chdir(test1_dir)
+
+        test_path = mbed_greentea_cli.get_local_host_tests_dir("")
+        self.assertEqual(test_path, "./test/host_tests")
+
+        os.chdir(curr_dir)
```
theotherjimmy: Why `os.chdir` here? Why not pass the `test1_dir` to `get_local_host_tests_dir`?

mbed_greentea/mbed_report_api.py
scartmell-arm: It was failing and I couldn't figure out a decent way of making it work for both python versions, I suppose I could revisit it.

test['single_test_output'] is a byte array in python2 and a string in python3 (which doesn't support encoding/decoding

mbed_greentea/mbed_target_info.py
scartmell-arm: As far as I could tell, it can't throw, the keys are checked in the if statements before and split always seems to return a list with a value, even if it's an empty string

theotherjimmy: @scartmell-arm It's more pythonic (and slightly faster) to just go for the indexing and catch the exception.

test/mbed_gt_cmake_handlers.py
scartmell-arm: Ah, you're right, I'll change that. I forgot to test with python 3 before making the PR so I ended up going through adding fixes to things that were breaking the CI.

mbed_greentea/mbed_target_info.py
scartmell-arm: @theotherjimmy Okay, in that case I'll change it to what you suggest somewhere else in the comments

mbed_greentea/mbed_test_api.py
scartmell-arm: I may have forgotten to remove that when I was cleaning up.

However, as with main_cli and the other top-level commands, it doesn't seem like a reasonable function to cover with unit testing; It takes a lot of arguments and aggregates a number of actually testable function calls.

Whilst it would be fairly trivial to cover all the lines in testing, that might be slightly disingenuous because covering all the branches/behaviour would be a lot of work and far from simple.

theotherjimmy: Would it also be slightly disingenuous to remove coverage from them?

mbed_greentea/mbed_report_api.py
bridadan: AH, well thanks for fixing it then 👍

test/mbed_gt_cmake_handlers.py
bridadan: No worries, I just learned about Python 2's `io` package a couple days ago myself!

test/mbed_gt_cli.py
scartmell-arm: That tests a different branch which is covered by the previous test. Passing an empty string asserts that the function checks that `./test/host_tests` exists.

mbed_greentea/mbed_test_api.py
scartmell-arm: In a way. But, I'd argue that it *shouldn't* be tested, rather than *hasn't* been tested, which excluding it would indicate clearer.

theotherjimmy: > I'd argue that it shouldn't be tested

I would not argue that.

> than hasn't been tested

This is definitely true.

I'd rather see these lines as red in a coverage report, rather than gray. That's just my opinion.

mbed_greentea/mbed_target_info.py
scartmell-arm: > I think this may throw an exception if keyword.split(':') does not have 2 elements.

It will only throw if the keyword string doesn't contain a colon. The previous line ensures that it will contain one.

> Should we be skipping these incorrectly formatted entries?

Maybe? They aren't necessarily incorrectly formatted, it's just checking that the mbed-target value exists and that it matches the input argument.

test/mbed_gt_cmake_handlers.py
scartmell-arm: Annoyingly, the types they use in each version don't seem to compatible for this case

```
Python 3.6.2 (v3.6.2:5fd33b5, Jul  8 2017, 04:57:36) [MSC v.1900 64 bit (AMD64)] on win32
>>> from io import StringIO
>>> print(type(StringIO().getvalue()))
<class 'str'>

Python 2.7.13 (v2.7.13:a06454b1afa1, Dec 17 2016, 20:53:40) [MSC v.1500 64 bit (AMD64)] on win32
>>> from io import StringIO
>>> print(type(StringIO().getvalue()))
<type 'unicode'>
>>>
>>> sys.stdout = out = StringIO()
>>> print("test")
  File "<stdin>", line 1, in <module>
TypeError: unicode argument expected, got 'str'

```

`six.StringIO` aliases compatible libraries for each version so I'll use that to avoid the try.

mbed_greentea/mbed_target_info.py
theotherjimmy: >     It will only throw if the keyword string doesn't contain a colon. The previous line ensures that it will contain one.

Nope:
```python
>>> a,b = "::".split(":")
  File "<stdin>", line 1, in <module>
ValueError: too many values to unpack (expected 2)
```

> They aren't necessarily incorrectly formatted,

This implies that split is the wrong thing to use here. How about:
```python
target, _, name = keyword.partition(':')
```
That can't raise an exception, ever:
```python
>>> "".partition(":")
('', '', '')
>>> ":".partition(":")
('', ':', '')
>>> "::::::".partition(":")
('', ':', ':::::')
````

File: test/mbed_gt_cli.py (hunk 11 above)
Comment (theotherjimmy) [reply test/mbed_gt_cli.py]: Thanks for the explanation. 👍

File: mbed_greentea/mbed_report_api.py [hunk 12]
```diff
@@ -218,10 +224,12 @@ def exporter_testcase_junit(test_result_ext, test_suite_properties=None):
                 test_cases.append(tc)
 
             ts_name = target_name
-            if test_suite_properties and target_name in test_suite_properties:
-                test_build_properties = test_suite_properties[target_name] 
+
+            if test_suite_properties is not None:
+                test_build_properties = test_suite_properties[target_name] if target_name in test_suite_properties else None
```
theotherjimmy: This could be:
```python
test_build_properties = test_suite_properties.get(target_name, None)
```

mbed_greentea/mbed_target_info.py
```diff
@@ -304,8 +309,8 @@ def add_target_info_mapping(mbed_classic_name, map_platform_to_yt_target=None, u
         yotta_target_search = list(set(yotta_target_search))    # Reduce repeated values
 
     # Add extra targets to already existing and detected in the system platforms
-    if mbed_classic_name in map_platform_to_yt_target:
-        yotta_target_search = list(set(yotta_target_search + map_platform_to_yt_target[mbed_classic_name]))
+    if map_platform_to_yt_target and mbed_classic_name in map_platform_to_yt_target:
+            yotta_target_search = list(set(yotta_target_search + map_platform_to_yt_target[mbed_classic_name]))
```
theotherjimmy: It looks like this got double indented. Could you fix the formatting?

theotherjimmy:
@bridadan I'm tempted to just merge it as is and get these other fixes in other small PRs. What do you think?

bridadan:
@theotherjimmy based on yours and @scartmell-arm's feedback, nothing seems too out of place here. Feel free to merge and patch up the odds and ends if you like. Just be sure to get it in a state your happy with before releasing 👍

PR #242, opened 2017-09-29
Remove traceback from parsing invalid keywords

theotherjimmy:


theotherjimmy:
@scartmell-arm @bridadan Could you review?

bridadan:
Do we need a test case for the case where it would return None?

theotherjimmy:
The test case does test for returning `None`. Are you looking for a specific test case?

bridadan:
AH yes, sorry, I was only looking at the diff.

Specifically I was wondering about the case when `keywords` and `name` are not in the JSON blob.

bridadan:
Not sure if that's a specific case we need to enforce that it always works with a test or not.

theotherjimmy:
That's a good test case to have. I'll write it in a moment.

scartmell-arm:
Seems like a reasonable change.

theotherjimmy:
** Facepalm **

I should write these tests.

theotherjimmy:
@bridadan Your tests have arrived!

PR #244, opened 2017-10-03
Fix remaining issues outlined in test coverage PR

scartmell-arm:
- Remove double indention
- Use get method with default value


theotherjimmy:
Thanks!

PR #248, opened 2017-10-19
Remove NUCLEO_F401RE

bcostm:
Due to the copy_method it is impossible to run any mbedOS5 tests on this platform.

### DIFF
--- a/mbed_greentea/mbed_target_info.py
+++ b/mbed_greentea/mbed_target_info.py
@@ -56,20 +56,6 @@ TARGET_INFO_MAPPING = {
                 "program_cycle_s": 4
             }
     },
-    "NUCLEO_F401RE" : {
-        "yotta_targets": [
-                {
-                    "yotta_target": "st-nucleo-f401re-gcc",
-                    "mbed_toolchain": "GCC_ARM"
-                }
-             ],
-        "properties" : {
-                "binary_type": ".bin",
-                "copy_method": "cp",
-                "reset_method": "default",
-                "program_cycle_s": 4
-            }
-        },
     "NRF51_DK" : {
         "yotta_targets": [
                 {


jeromecoutant:
Hi
Please, release a new greentea version as soon as this patch is merged.
Thx

theotherjimmy:
Hey guys, Thanks for the PR. Would it be possible to change the `copy_method` instead of removing the target?

bcostm:
Ok it is possible but why keeping only this Nucleo board ? Is it to give an example ?

theotherjimmy:
Oh, I was just trying to avoid removing targets in general, if they can be fixed. I'm pretty new to greentea; Did we remove other Nucleo boards the past?

bridadan:
These are old yotta definitions. When mbed OS was distributed using yotta, greentea used to hardcode platform information in the tool. These definitions really aren't used anymore as far as I know.

@bcostm Was this a recent regression that was released with Greentea? I'm pretty sure I've used this platform with Greentea and ran mbed OS 5 tests in the recent past. Are you getting a specific error?

jeromecoutant:
We can make tests with greentea v1.2.5

theotherjimmy:
@bridadan Thanks for the clarification.

bridadan:
@jeromecoutant @bcostm if it's broken, can please verify which version is broken so we can reproduce the issue here?

jeromecoutant:
Issue came with this:
https://github.com/ARMmbed/greentea/pull/225

bridadan:
AH great, thanks @jeromecoutant! That makes sense, oops.

@theotherjimmy This is ok to come in as a quick fix if you want. We should really to start to look at it either stripping out legacy yotta stuff OR figure out a better way for them to coexist.

bcostm:
Thanks for the correction but I really don't see the purpose to keep NUCLEO_F401RE in this file ? Because the copy method is the same for all ST boards, so no particular reason for this board (excepted if I miss something?)

theotherjimmy:
> Thanks for the correction but I really don't see the purpose to keep NUCLEO_F401RE in this file ? 

Removing the NUCLEO_F401RE from this file would technically be a breaking change. That's why I opted to correct the bug that allowed this to override the targets.json definition.

> Because the copy method is the same for all ST boards, so no particular reason for this board (excepted if I miss something?)

This way I can do a patch release, and not a major release.

----------------

I would like to remove this yotta stuff for version 2.0.0.

bridadan:
@bcostm We intend to remove all of the boards in that file, not just the NUCLEO_F401RE for version 2.0.0

jeromecoutant:
Hi
When this v2 is planned ?
Thx

bridadan:
Hi Jerome,

We don't currently have a plan when v2 will be released, we're still trying to figure out all the changes that would take place. If you're waiting on specific changes please be sure to let us know so we can get them in sooner.

PR #250, opened 2017-10-20
Correct target property priority

### LINKED ISSUE #249: Target property priority incorrect
Currently we have priority as follows:
```
internal yotta blob > targets.json > tool default
```
This is a bug.

Instead the priority should be:
```
targets.json /w default > internal yotta blob > tool delaut
```

This implies a few test cases:

In targets.json        | In yotta blob | property used | Currently Works
---------------------- | ------------- | ---------------- | ---------------
Yes, with property | No            | `targets.json`   | Yes
Yes, without property| No            | default   | Yes
Yes, with property | Yes           | `targets.json`   | No
Yes, without property | Yes           | default   | No
No                     | No           | default       | Yes
No                     | Yes           | yotta blob       | Yes

@bridadan Is this the issue masked by #248?

theotherjimmy:
And refactor into a bunch more tests.

Fixes #249
Closes #248

### DIFF
--- a/mbed_greentea/mbed_target_info.py
+++ b/mbed_greentea/mbed_target_info.py
@@ -383,25 +383,25 @@ def get_platform_property(platform, property):
     :return: property value, None if property not found
     """
 
-    default = get_platform_property_from_default(property)
-    from_targets_json = get_platform_property_from_targets(
+    default = _get_platform_property_from_default(property)
+    from_targets_json = _get_platform_property_from_targets(
         platform, property, default)
     if from_targets_json:
         return from_targets_json
-    from_info_mapping = get_platform_property_from_info_mapping(platform, property)
+    from_info_mapping = _get_platform_property_from_info_mapping(platform, property)
     if from_info_mapping:
         return from_info_mapping
     return default
 
-def get_platform_property_from_default(property):
+def _get_platform_property_from_default(property):
     with suppress(KeyError):
         return TARGET_INFO_MAPPING['default'][property]
 
-def get_platform_property_from_info_mapping(platform, property):
+def _get_platform_property_from_info_mapping(platform, property):
     with suppress(KeyError):
         return TARGET_INFO_MAPPING[platform]['properties'][property]
 
-def platform_property_from_targets_json(targets, platform, property, default):
+def _platform_property_from_targets_json(targets, platform, property, default):
     """! Get a platforms's property from the target data structure in
             targets.json. Takes into account target inheritance.
     @param targets Data structure parsed from targets.json
@@ -415,7 +415,7 @@ def platform_property_from_targets_json(targets, platform, property, default):
         return targets[platform][property]
     with suppress(KeyError):
         for inherited_target in targets[platform]['inherits']:
-            result = platform_property_from_targets_json(targets, inherited_target, property, None)
+            result = _platform_property_from_targets_json(targets, inherited_target, property, None)
             if result:
                 return result
     if platform in targets:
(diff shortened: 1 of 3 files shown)


theotherjimmy:
This is actually more - than +, if you exclude tests.

theotherjimmy:
lol, coverage increased but testing failed

theotherjimmy:
Double coverage!

mbed_greentea/mbed_target_info.py
```diff
@@ -381,82 +392,65 @@ def get_platform_property(platform, property):
...
-
-    # Check if info is available for a specific platform
-    if platform in TARGET_INFO_MAPPING:
-        if property in TARGET_INFO_MAPPING[platform]['properties']:
-            return TARGET_INFO_MAPPING[platform]['properties'][property]
+    default = _get_platform_property_from_default(property)
+    from_targets_json = _get_platform_property_from_targets(
+        platform, property, default)
+    if from_targets_json or not default:
+        return from_targets_json
```
bridadan: Just wanna make sure I understand this logic. If target is not in targets.json, `from_targets_json` should always return `None`. If  for some reason `default` isn't set, then this should return `None` (the value of `from_targets_json`). That makes sense to me, I just want to make sure I understand this block since the logic wasn't immediately apparent to me when reading the code.

setup.py
```diff
@@ -50,13 +50,16 @@ def read(fname):
...
-        "mock",
-        "colorama>=0.3,<0.4"])
+                        "PySerial>=3.0",
+                        "mbed-host-tests>=1.2.0",
+                        "mbed-ls>=1.2.15",
+                        "junit-xml",
+                        "lockfile",
+                        "mock",
+                        "six",
+                        "colorama>=0.3,<0.4"],
```
bridadan: Extra comma on the end here, is that ok? Can't remember how Python treats that.

test/mbed_gt_target_info.py
```diff
@@ -338,8 +340,168 @@ def test_parse_add_target_info_mapping(self):
...
+    # Num | In targets.json | In yotta blob | In Default | property used
+    # --- | --------------- | ------------- | ---------- | --------------
+    # 1   | Yes             | No            | Yes        |`targets.json`
+    # 2   | Yes             | Yes           | Yes        |`targets.json`
+    # 3   | No              | Yes           | Yes        | yotta blob
+    # 4   | No              | No            | Yes        | default
+    # 5   | No              | No            | No         | None
+    # 6   | Yes             | No            | No         |`targets.json`
+    # 7   | Yes             | Yes           | No         |`targets.json`
+    # 8   | No              | Yes           | No         | None
```
bridadan: Case #8 doesn't seem right. Should it not use the value in the yotta blob?

mbed_greentea/mbed_target_info.py
theotherjimmy: That's correct, and the case 8 below.

setup.py
theotherjimmy: It's cool with it.

theotherjimmy:
@bridadan I basically removed tho commit that changed how case 8 works. It should be "yotta" now.

setup.py
theotherjimmy: Eh, I'm rebasing anyway. I'll remove it.

theotherjimmy:
Coveralls, calm down!

theotherjimmy:
@bridadan Did I address all of your comments?

jeromecoutant:
Hi
Could you please release a new greentea version ?
Thx

theotherjimmy:
Yeah, sure. Give me a few minutes.

theotherjimmy:
Release punished to github and pip. @jeromecoutant Let me know if you need something else.

Issue #252, opened 2017-11-15
GT still using mbedls.lstools_base (deprecated)


cmonr:
Needs to be updated with list_mbeds function call.

theotherjimmy:
Looks like this one is in htrun.

cmonr:
Merged.

PR #253, opened 2017-11-16
Updates deprecated mbedls API call 

cmonr:
Reference: https://github.com/ARMmbed/greentea/issues/252


bridadan:
Thanks for the patch @cmonr ! This looks good, can you also update the dependency of mbed-ls: https://github.com/ARMmbed/greentea/blob/master/setup.py#L58

You'll want to change it to `mbed-ls>=1.3.4`

PR #257, opened 2018-01-05
Accept --retry-count argument that retries on test failure

studavekar:
introduced --retry-count arguments to specify retry count of individual test
failure. By default, there is no retry.

### DIFF
--- a/mbed_greentea/mbed_test_api.py
+++ b/mbed_greentea/mbed_test_api.py
@@ -103,6 +103,42 @@ def get_test_result(output):
     return TEST_RESULT_TIMEOUT
 
 
+def run_command(cmd):
+    """! Runs command and prints proc stdout on screen
+    @paran cmd List with command line to execute e.g. ['ls', '-l]
+    @return Value returned by subprocess.Popen, if failed return None
+    """
+    try:
+        p = Popen(cmd,
+                  stdout=PIPE,
+                  stderr=STDOUT)
+    except OSError as e:
+        gt_logger.gt_log_err("run_host_test.run_command(%s) failed!" % str(cmd))
+        gt_logger.gt_log_tab(str(e))
+        return None
+    return p
+
+def run_htrun(cmd, verbose):
+    # detect overflow when running tests
+    htrun_output = str()
+    # run_command will return None if process can't be opened (Issue #134)
+    p = run_command(cmd)
+    if not p:
+        # int value > 0 notifies caller that starting of host test process failed
+        return RUN_HOST_TEST_POPEN_ERROR
+
+    for line in iter(p.stdout.readline, b''):
+        htrun_output += line
+        # When dumping output to file both \r and \n will be a new line
+        # To avoid this "extra new-line" we only use \n at the end
+        if verbose:
+            sys.stdout.write(line.rstrip() + '\n')
+            sys.stdout.flush()
+
+    # Check if process was terminated by signal
+    returncode = p.wait()
+    return returncode, htrun_output
+
 def run_host_test(image_path,
                   disk,
                   port,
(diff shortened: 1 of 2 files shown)


theotherjimmy:
Job failure was related to github outage. Restarting.

theotherjimmy:
Note to self (@theotherjimmy) Bump the minor version after this PR

theotherjimmy:
@studavekar Could we add a unit test for this? Something such as mocking a failing then passing test and verifying that this correctly implements the retry would be sufficient. Let me know if you would like any assistance with the test-writing.

bridadan:
> Note to self (@theotherjimmy) Bump the minor version after this PR

Maybe we should add some labels for this kind of thing? Just to keep ourselves sane.

Also @studavekar I know you and I have talked about this in the past. Generally we try not to run a retry if a test actually fails because we don't want to mask an actual bug if possible. Is there a specific use case that is being used for? Can the retry for the test be inside the test case code itself?

jupe:
I’ve same feeling; we might actually hide some bug or unstable test if we retry failing tests, especially if this is activated in CI. I would also like to understand original problem this tries to fix. I can imagine that this could help some instability problems in test environment but even then those problems should be raised as a issue somewhere rather than hide them. Anyway - implementation point of view looks good.

theotherjimmy:
@bridadan We are now using the enhancement label for tracking minor version bumps
-----------------
@jupe Thanks for expanding on the "should we" issue.

studavekar:
@bridadan @jupe 

yes concerns 100 % valid having a re-try may hide valid bugs, Below are the reason for adding this option at first place.

- infrastructure issues causing tests to fail - device flash, one device having a bad connection, intermittent network issue. - now with RaaS have this test run on other devices without making entire matrix fail. - Which saved us a lot of time in order of hours and also some frustration of false failure because of infra. 

- Tool software issue which is marked an can't be fixed or can't be reproduced - like Daplink programming failure, UART packets been corrupted, ST link issues. 

- In CI if we retry X times and it fails - it gives higher confidence that there is a genuine failure and not an intermittent failure - 99 % of the time for an intermittent failure test is re-triggered it passes so the issue is never debugged or looked till the failure rate is high.

- Finally, by default, there are no retries :) 

Should we use if by default in CI system - is open for discussion 
Should we have such option - I feel we should most of the test tools provide an option with a retry.

bridadan:
> In CI if we retry X times and it fails - it gives higher confidence that there is a genuine failure and not an intermittent failure - 99 % of the time for an intermittent failure test is re-triggered it passes so the issue is never debugged or looked till the failure rate is high.

That's a good point, it really doesn't actually change our work flow in general. So it basically provides the option to be sure something is really failing.

It'd be nice in the report if we could have a summary showing how many retries were attempted just so it doesn't slip under the radar. But if someone is monitoring this already then I suppose that isn't that big of a deal for me.

> Should we use if by default in CI system - is open for discussion
> Should we have such option - I feel we should most of the test tools provide an option with a retry.

Agreed, I actually don't have a problem for this being an option in Greentea. Whether or not y'all use it in CI systems is really the test and OS team's call, not Greentea.

mbed_greentea/mbed_test_api.py
```diff
@@ -165,6 +166,27 @@ def run_command(cmd):
             return None
         return p
 
+    def run_htrun(cmd):
```
theotherjimmy: Could you pull this and `run_command` out of this function?

studavekar: sure.

theotherjimmy:
Coverage decrease is related to the fact that coverage was always that bad, but masked by poor style (the function-in-a-function thing).

PR #258, opened 2018-01-05
Handle uart corruption in report generation

studavekar:
Its been seen that when running devices in parallel about 20 devices,sometime
uart buffer received by greentea is corrupted. In case where tags like
`__testcase_start` corrupted `result_test_cases[testcase_id]['utest_log']` is
not set and will results in exception when generating report.

This change will make sure we have reports generated for the run.


theotherjimmy:
@studavekar Could we handle this by assigning the `"uart corrupted"` string at the beginning, and filling in the string with the correct uart log when we have it? This way exceptions that we may throw won't cause the same breakage down the road.

Further, Could we have a test for this behavoir please? Something simple like mocking a corrupt UART is sufficient. Let me know if you would like assistance writing the test case.

bridadan:
Just a nit from @theotherjimmy's comment: Let's not say `uart corrupted` since there are plans to have testing over other channels in the future.  Just go with `device output corrupted` or something.

theotherjimmy:
@bridadan Yes. Thanks for the correction.

studavekar:
> @studavekar Could we handle this by assigning the "uart corrupted" string at the beginning, and filling in the string with the correct uart log when we have it? This way exceptions that we may throw won't cause the same breakage down the road.

Have updated to assign utest_log if it's not already defined, also have string "__testcase_start tag not found"

PR #263, opened 2018-02-15
Fix test name globing

### LINKED ISSUE #262: Test names are not correctly globbed
Test names only respect a wildcard that is placed at the end of the string. Ex. "mbed-os-*".

However, it does not respect the wildcard anywhere else. Ex. "*-timer"

The build tools accept these wildcards, so greentea should as well. This is the line responsible: https://github.com/ARMmbed/greentea/blob/32b95b44be653c3db527c02e1c5e1ffdc7d37f6f/mbed_greentea/mbed_greentea_cli.py#L146

Should be switched to `fnmatch`.

(This is mostly a note to myself to fix it)

bridadan:
This fixes #262. It also fixes testing on Windows since directory separators are silly.

### DIFF
--- a/mbed_greentea/mbed_greentea_cli.py
+++ b/mbed_greentea/mbed_greentea_cli.py
@@ -23,6 +23,7 @@ import os
 import sys
 import random
 import optparse
+import fnmatch
 from time import time
 try:
     from Queue import Queue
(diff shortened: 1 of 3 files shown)


mbed_greentea/mbed_greentea_cli.py
```diff
@@ -142,18 +131,17 @@ def filter_names_by_prefix(test_case_name_list, prefix_name):
         test_list = test_by_names.split(',')
         gt_logger.gt_log("test case filter (specified with -n option)")
 
+
```
theotherjimmy: Could you remove this whitespace change

bridadan: I refuuuuuuse!

(yeah no problem)

PR #266, opened 2018-03-20
set timeout to 120 sec for GRM

studavekar:
The default timeout is 60 secs bump it to 120 secs to handle flash retry.


juhaylinen:
Any update on this?
Default timeout may not be enough when flashing large images. Please add polling-timeout to CLI options.

theotherjimmy:
@juhaylinen Could you point to or open an issue for this "polling-timeout" that you want?

juhaylinen:
@theotherjimmy issue created https://github.com/ARMmbed/greentea/issues/269

jupe:
Is this timeout for what part ? only for flashing ? This seems to be needed because sometimes current timeout is too short causing unexpected test failures.

Jekaah:
Seems problems that we are facing in Update-Client CI tests are also related to this - flashing timeouts. 

It would be very important to get this fix forward. Is there any other workaround that could be used?

cC: @evedon @bogdanm @LiyouZhou

bridadan:
It's fine with me, the only thing I think we should address is adding a greentea argument for the option @studavekar is adding in this PR.

studavekar:
> It's fine with me, the only thing I think we should address is adding a greentea argument for the option @studavekar is adding in this PR.

Yeah will add an argument.

studavekar:
Now as this comes as arguments have set the default value to 60 secs to which in inline with htrun. One may tweak this value as per the requirement.

studavekar:
@theotherjimmy can you please review it?

theotherjimmy:
Oh another reason to remove the "Call it on the command line" problem.

PR #270, opened 2018-05-11
add "--fm" option to support FastModels testing in greentea

jamesbeyond:
# Description

In order to support FastModels in mbed testing framework, an extra option "--fm" been added to greentea.

This option will allow greentea hooks up with fm_agent module, and run tests over FastModels

The command line format of the --fm is following the RaaS_Client "--grm" convention

# Details
* "--fm" been added to mbedgt
* user could use __mbedgt  --fm \<model_name>:<config_name>__
  * e.g.  __mbedgt  --srm FVP_MSP2_M3:DEFAULT__

This PR is related with other PRs regarding to FastModel support in testing framework 
ARMmbed/mbed-ls#345
ARMmbed/htrun#187

CC. @theotherjimmy @studavekar @bridadan dan @bulislaw

### DIFF
--- a/README.md
+++ b/README.md
@@ -578,6 +578,21 @@ $ mbedls
 ```
 In this case, you won't test one target, the LPC1768.
 
+### Testing on Fast Model FVPs
+
+Fast Models FVPs are software models for Arm reference design platfrom
+
+Greentea supports running test on Fast Models. And [mbed-fastmodel-agent](https://github.com/ARMmbed/mbed-fastmodel-agent) module is required for this purpose. 
+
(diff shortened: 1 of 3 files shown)


bridadan:
Just a heads up, `imp` has been deprecated in Python 3 since 3.4: https://docs.python.org/3/library/imp.html. We may want to address that in the future. But hooray for fastmodels coming in! 🎉

jamesbeyond:
@bridadan thanks for reminding, I was aware of that. 
when mbed tools moving python 3.x, `imp` will be changed to `importlib`

PR #271, opened 2018-06-01
Fixes string issues with running in Py3

cmonr:
Small fixes to allow greentea to run with Python 3.


cmonr:
@bridadan  👍

cmonr:
@theotherjimmy @bridadan I would hold off on doing a release of the toll until early next week. I suspect I'm still going to find more issues this weekend.

PR #272, opened 2018-06-04
Update README.md

JanneKiiskila:
Add the sudo -H you typically need plus update instructions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,13 @@ $ mbedgt --version
 You can also install it manually via pip.
 
 ```
-pip install mbed-greentea
+sudo -H pip install mbed-greentea
+```
+
+You can update an existing installation of Greentea via pip, too.
+
+```
+sudo -H pip install -U mbed-greentea
 ```
 
 ## Test specification JSON format


theotherjimmy:
@JanneKiiskila This is not always true, -H is not needed at all. Many users may insist on installing greentea in a virtualenv, or as a user package. Instead of assuming that everyone installs pip packages globally, could we add a note about install location and that you may need sudo, or --user, or something else.

README.md
```diff
@@ -32,7 +32,13 @@ $ mbedgt --version
 You can also install it manually via pip.
 
 ```
-pip install mbed-greentea
+sudo -H pip install mbed-greentea
```
theotherjimmy: `sudo` is not available on every platform. Could you add instructions for platforms that don't have `sudo`? Could you explain why `-H` is used?

bridadan:
For the reasons stated by @theotherjimmy, we should be guiding users towards options such as `pip install --user mbed-greentea` or using virtual environments. Would you like to update this PR? If not we can close this and add it in a follow up.

bridadan:
Hi @JanneKiiskila , we are moving the project today, so I'll be closing this issue. If this is an issue you are still interested in, please go ahead and reopen this issue! I will then transfer it to the https://github.com/armmbed/mbed-os-tools repo. Thanks!

PR #273, opened 2018-06-04
Release Version 1.4.0

theotherjimmy:
# Features
  * expose polling timeout arguments -P
  * Detect assert failures in htrun messages and log as error
  * Add __main__ file to simplify running from source
  * allow skipping of tests using * expansion

 # Fixes
  * Fix string issues with running in Py3


theotherjimmy:
@jupe As requested.

PR #278, opened 2018-07-19
add try/exception to handle UnicodeDecodeError

aashishc1988:
Due to lack of pull-up on different parts, the re-initialization of uart causes it to send an invalid character which we are not able to handle properly on greentea side. While partners are investigating the issue and fix, we will add this change in to handle the exception gracefully, as this bug is blocking the addition of multiple parts to our CI.

### DIFF
--- a/mbed_greentea/mbed_test_api.py
+++ b/mbed_greentea/mbed_test_api.py
@@ -134,7 +134,7 @@ def run_htrun(cmd, verbose):
         try:
             htrun_output += line.decode('utf-8')
         except UnicodeDecodeError:
-            pass
+            gt_logger.gt_log_err("UnicodeDecodeError encountered!")
         # When dumping output to file both \r and \n will be a new line
         # To avoid this "extra new-line" we only use \n at the end
 
@@ -146,7 +146,7 @@ def run_htrun(cmd, verbose):
             try:
                 sys.stdout.write(line.decode('utf-8').rstrip() + '\n')
             except UnicodeDecodeError:
-                pass
+                gt_logger.gt_log_err("UnicodeDecodeError encountered!")
             sys.stdout.flush()
 
     # Check if process was terminated by signal
--- a/mbed_greentea/mbed_test_api.py
+++ b/mbed_greentea/mbed_test_api.py
@@ -131,7 +131,10 @@ def run_htrun(cmd, verbose):
     htrun_failure_line = re.compile('\[RXD\] (:\d+::FAIL: .*)')
 
     for line in iter(p.stdout.readline, b''):
-        htrun_output += line.decode('utf-8')
+        try:
+            htrun_output += line.decode('utf-8')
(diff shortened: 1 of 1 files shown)


mbed_greentea/mbed_test_api.py
```diff
@@ -140,7 +143,10 @@ def run_htrun(cmd, verbose):
             gt_logger.gt_log_err(test_error.group(1))
 
         if verbose:
-            sys.stdout.write(line.decode('utf-8').rstrip() + '\n')
+            try:
+                sys.stdout.write(line.decode('utf-8').rstrip() + '\n')
```
studavekar: can we replace `os.linesep` instead of `\n`

aashishc1988:
For now, if the decode fails no output will be logged. Last we discussed, we need to figure out a way to log that as an error and the test should fail, but we would do that once the partners have fixed the issue on there side so that it doesnt block the integration of boards to CI. I can make a note of this in the tracking jira as well.

aashishc1988:
We can add some sort of log when these errors occur so we can use it later to analyze what devices are still showing the behavior. I will add some sort of meaningful log.

mbed_greentea/mbed_test_api.py
aashishc1988: I agree. We should file a different PR for that

theotherjimmy:
@aashishc1988 Would it make more sense to use a different option for that decode? possibly to decode with replacement or decode ignoring errors (`errors="replace"` or `errors="ignore"`) https://docs.python.org/3/library/stdtypes.html?highlight=decode#bytes.decode

We could also force the output to remain bytes for the duration of this command, passing handling of encoding errors to the next level up.

aashishc1988:
@theotherjimmy good suggestion regarding different option for decode. However, if we intend to print some sort of meaningful message if/when the error occurs, adding a try/catch seems clean and simple, unless I am missing some other decode option that takes care of that. 
Regarding the second suggestion, please note that this is a temporary solution to get around this problem, https://github.com/ARMmbed/mbed-os/issues/7511, and we would be actively tracking that issue to make sure we revert these changes once that is fixed, so ideally it would be nice to limit the scope of this solution to one particular module. Thoughts?

theotherjimmy:
I'm not keen to release a version of greentea that contains a "not our final solution, but a workaround", as changing the workaround would constitute a breaking change. Further, I'm not convinced that changing the types from str-> bytes would be a change that needs to be made anywhere else.

theotherjimmy:
@aashishc1988 Could we change the `htrun_failure_line` regex to be a bytes-pattern instead? then we could avoid decoding entirely.

aashishc1988:
Come to think of it, this PR is not even needed. On htrun, we are now going to ignore the non-unicode characters and not add them to stream buffer, so on GT side, those are not even encountered. Sorry I missed that earlier. We can close this PR.

PR #282, opened 2018-09-28
set forced_reset_timeout 10 sec for RAPIDIOT_K64F

studavekar:
RAPIDIOT_K64F devices take a longer time to boot up.

### DIFF
--- a/mbed_greentea/mbed_target_info.py
+++ b/mbed_greentea/mbed_target_info.py
@@ -69,7 +69,7 @@ TARGET_INFO_MAPPING = {
     },
     "RAPIDIOT_K64F" : {
         "properties" : {
-                "forced_reset_timeout":10
+                "forced_reset_timeout":7
             }
     },
     "NUCLEO_F401RE" : {
--- a/mbed_greentea/mbed_target_info.py
+++ b/mbed_greentea/mbed_target_info.py
@@ -67,6 +67,11 @@ TARGET_INFO_MAPPING = {
                 "program_cycle_s": 4
             }
     },
+    "RAPIDIOT_K64F" : {
+        "properties" : {
+                "forced_reset_timeout":10
+            }
+    },
     "NUCLEO_F401RE" : {
         "yotta_targets": [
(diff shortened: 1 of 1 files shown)


studavekar:
cc: @maclobdell

maclobdell:
cc @mmahadevan108 
please review.  
please note that this requires updating your greentea version used for testing.  we will do a separate pr for updating mbed-os to require the latest greentea (and other test related) python modules.

bridadan:
@studavekar @maclobdell Is the increased boot time due to the fact that the interface chip queries the MCU first? Htrun currently sets a default `forced_reset_timeout` of 1 second. 10 seconds is an order of magnitude increase, so it'd be good to make sure its absolutely necessary!

maclobdell:
@bridadan - the rapid Iot platform is unique.  It has a bootloader that interrogates the contents of a spi flash to check if a newer application is available for each processor (K41Z & K64F).  This takes a few seconds.  I agree that 10 seconds might be too long to wait and could blow up the test time to epic proportions of there is some fundamental problem.  @mmahadevan108 - Can you suggest a more appropriate timeout for this target?  Will it be guaranteed to boot the application within 3 sec, 4 sec, 5 sec?

mmahadevan108:
Thank you, this change will work for RapidIoT

theotherjimmy:
@mmahadevan108, @maclobdell was asking if 10 sec is really required as this will greatly slow down the Mbed OS CI. Each test will wait an additional 10 seconds, which especially harms tests that take <1 sec (a good portion of the tests). Our CI runtime is predicted by the longest time it takes for a platform to run all of it's tests, so we're wondering if we can safely save a few seconds * number of tests (100s) for this platform.

mmahadevan108:
From my testing, forced_reset_timeout of 5 secs did not work. 
6 secs upwards worked. I would suggest we go with 7 secs.

studavekar:
@mmahadevan108 thanks will set it to 7 secs

Issue #288, opened 2018-11-10
mbed-greentea 1.4.0 failed on Japanese Windows 10


toyowata:
Our partner Seeed reported this and I also reproduced the issue.
cc @ytsuboi

Problem:
`mbed test` failed by UnicodeDecodeError on Windows 10.

[python version]
Python 2.7.15

[mbed-os version]
mbed-os (#bf6f2c3c6434, tags: latest, mbed-os-5.10.3)

[Windows version]
Microsoft Windows [Version 10.0.17134.345]

Steps to reproduce:
* Import mbed-os-5.10.3
* Install missing python modules
* Run mbed test (any target, any toolchain)

```
PS C:\test\mbed-os> mbed test -m DISCO_L475VG_IOT01A -t gcc_arm -n tests-mbed_hal-ticker --run
[mbed] Auto-installing missing Python modules...
mbedgt: greentea test automation tool ver. 1.4.0
mbedgt: test specification file '.\BUILD\tests\DISCO_L475VG_IOT01A\GCC_ARM\test_spec.json' (specified with --test-spec option)
mbedgt: using '.\BUILD\tests\DISCO_L475VG_IOT01A\GCC_ARM\test_spec.json' from current directory!
mbedgt: detecting connected mbed-enabled devices...
mbedgt: detected 1 device
mbedgt: processing target 'DISCO_L475VG_IOT01A' toolchain 'GCC_ARM' compatible platforms... (note: switch set to --parallel 1)
mbedgt: test case filter (specified with -n option)
        tests-mbed_hal-ticker
        test filtered in 'tests-mbed_hal-ticker'
mbedgt: running 1 test for platform 'DISCO_L475VG_IOT01A' and toolchain 'GCC_ARM'
mbedgt: mbed-host-test-runner: started
Exception in thread Thread-3:
Traceback (most recent call last):
  File "c:\python27\lib\threading.py", line 801, in __bootstrap_inner
    self.run()
  File "c:\python27\lib\threading.py", line 754, in run
    self.__target(*self.__args, **self.__kwargs)
  File "c:\python27\lib\site-packages\mbed_greentea\mbed_greentea_cli.py", line 489, in run_test_thread
    verbose=verbose)
  File "c:\python27\lib\site-packages\mbed_greentea\mbed_test_api.py", line 318, in run_host_test
    returncode, htrun_output = run_htrun(cmd, verbose)
[...]

PR #289, opened 2018-11-14
Ignoring bad unicode characters

bridadan:
This should stop the exception occurring in #288.

### DIFF
--- a/mbed_greentea/mbed_test_api.py
+++ b/mbed_greentea/mbed_test_api.py
@@ -131,7 +131,8 @@ def run_htrun(cmd, verbose):
     htrun_failure_line = re.compile('\[RXD\] (:\d+::FAIL: .*)')
 
     for line in iter(p.stdout.readline, b''):
-        htrun_output += line.decode('utf-8')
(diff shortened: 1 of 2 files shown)


bridadan:
> We could also use the system default encoding, and not ignore.

I'm not as familiar with character encoding (in and out of Python) as I'd like to be, could explain that a bit more? Instead of specifying `utf-8` would it be more something like calling `sys.getdefaultencoding()` to get the right encoding?

theotherjimmy:
@bridadan Exactly correct. `sys.getdefaultencodinfg()` is the system default encoding.

bridadan:
Gotcha. Maybe a separate PR?

theotherjimmy:
@bridadan It would be an alternate solution to the same problem.

bridadan:
This PR makes the last few calls to `decode` that don't have `'ignore'` added consistent with the rest of the codebase. There other places in the codebase where we're currently doing things like `s.decode('utf-8', 'ignore')`, so I'd prefer to change all of those in one PR with some added manual testing.

theotherjimmy:
Sounds good to me.

PR #301, opened 2021-05-24
Standalone Greentea host

wernerlewis:
Development of greentea host was moved into [mbed-os-tools](https://github.com/ARMmbed/mbed-os-tools/tree/master/packages/mbed-greentea/mbed_greentea) and this repository was deprecated. Following the decision to make Greentea standalone again, this repository is reused.

The following changes are made:

- `mbed-os-tools` is not required, now using `mbedls` directly for device detection. `htrun` is to be included in this repo.
- Unused yotta files and references are removed.
- Mbed references are removed, package has not yet been renamed.
- Errors found with `black`, `mypy` and `flake8` are resolved.
- Docstrings are updated and standardised.


rwalton-arm (changes requested):
In the commit "Initial greentea host" we seem to be doing far too much. We should split this into separate commits:
1. Copies the code over as is
2. Reformats with black
3. Adds tox.ini
4. Removes mbed refs

It doesn't matter as much about every commit passing tests at this point as we don't have any CI enabled. At the moment it's too difficult to review any changes we've made.

We also need to work out how we integrate this PR and #302 - I would've expected one of the PRs to be based on the other if there's any crossover. If they merge cleanly without conflicts as-is then great.

wernerlewis:
I've rebased and updated so that this integrates on top of the previous work, and split out the commits.

When running `gt`, `htrun` passes a few test suites, but fails the majority. Replacing `htrun` with `mbedhtrun` to use the existing version from `mbed-os-tools` doesn't show the same issue.

This seems to be due to the host_tests not being loaded from `mbed-os`: [for example](https://github.com/ARMmbed/mbed-os/blob/master/rtos/tests/TESTS/host_tests/timing_drift_auto.py#L19-L23), due to the change in module name. So in `host_registry.py` host_tests aren't being registered, leading to "host test not detected" errors. This means `gt`/`htrun` will fail to run tests that rely on host_tests defined in `mbed-os` until those are updated.

src/greentea/gtea/greentea_cli.py
```diff
@@ -31,14 +36,17 @@ def get_local_host_tests_dir(path):
...
-    @ctest_test_list List of tests, derived from test specification
-    @test_by_names Command line switch -n <test_by_names>
-    @skip_test Command line switch -i <skip_test>
-    @param test_spec Test specification object loaded with --test-spec switch
-    @return
-    """
+    """Create a filtered test case list.
+
+    Args:
+        ctest_test_list: List of tests, derived from test specification.
```
rwalton-arm, line 42: `ctest_test_list` is an outdated name. The tool doesn't use CTestTestFile.cmake anymore. Let's just change this to `test_list`?

src/greentea/gtea/greentea_dlm.py
```diff
@@ -2,28 +2,33 @@
 # Copyright (c) 2021 Arm Limited and Contributors. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 #
-
+"""Implementation of greentea locking. Note: relies on deprecated package lockfile."""
```
rwalton-arm, line 5: What's the value of all this lockfile stuff? Why do we need it? This docstring would be the place to explain why the module exists at all.

src/greentea/gtea/greentea_dlm.py
```diff
@@ -32,22 +37,38 @@ def greentea_get_app_sem():
...
+    """
     greentea_home_dir_init()
     file_path = os.path.join(HOME_DIR, GREENTEA_HOME_DIR, GREENTEA_GLOBAL_LOCK)
     lock = lockfile.LockFile(file_path)
     return lock
 
 
 def greentea_update_kettle(greentea_uuid):
-    from time import gmtime, strftime
+    """Update greentea kettle.
```
rwalton-arm, line 67: What's a "greentea kettle"?

src/greentea/gtea/greentea_dlm.py
```diff
@@ -62,7 +83,11 @@ def greentea_update_kettle(greentea_uuid):
 
 
 def greentea_clean_kettle(greentea_uuid):
-    """Clean info in local file system config file"""
+    """Clean info in local file system config file.
```
rwalton-arm, line 86: What's a "local file system config file" and why is it called a "kettle"?

src/greentea/gtea/greentea_hooks.py
```diff
@@ -2,43 +2,38 @@
...
-
     cmd = None
 
     def __init__(self, name, cmd):
-        GreenteaTestHook.__init__(self, name)
+        """Initialise CLI Test Hook.
+
+        Args:
+            name: Name of the hook.
+            cmd: Command to be ran.
```
rwalton-arm, line 25: Shouldn't it be "command to be run"? Has the command already been executed at this point?

src/greentea/gtea/greentea_hooks.py
```diff
@@ -2,43 +2,38 @@
...
+        """
+        self.name = name
         self.cmd = cmd
 
     def run_cli_process(self, cmd):
```
rwalton-arm, line 30: `subprocess.run` makes this wrapper unnecessary.

src/greentea/gtea/greentea_hooks.py
```diff
@@ -138,55 +151,52 @@ def expand_parameters(expr, expandables, delimiter=" "):
 
 
 class LcovHook(GreenteaCliTestHook):
-    """! Class used to define a LCOV hook"""
+    """Class used to define a LCOV hook."""
```
rwalton-arm, line 154: "an LCOV hook"

src/greentea/gtea/greentea_log.py
```diff
@@ -14,7 +14,7 @@
 
 
 class GreenTeaSimpleLockLogger(object):
-    """Simple locking printing mechanism."""
+    """Simple lock printing mechanism."""
```
rwalton-arm, line 17: What's a lock printing mechanism? Do we mean a thread-safe printing mechanism? I assume we're trying to guarantee serialised output in a multi-threaded context with this object.

src/greentea/gtea/report_api.py
```diff
@@ -160,11 +174,14 @@ def exporter_testcase_text(test_result_ext):
...
+    """Export test results in JUnit XML compliant format.
+
+    Args:
+        test_result_ext: Extended report from Greentea.
+        test_spec: Dictionary of test build names to test suite properties.
+
+    Returns:
+        String containing Junit XML formatted test result output.
     """
     from junit_xml import TestSuite, TestCase
```
rwalton-arm, line 186: I think we should create a task to remove the function scoped imports: it's not good form to import a bunch of stuff in a public api function, especially if it's going to be called more than once (we'll run the import code again on each call). It's a small cost to import these things at file scope, and a cost we only need to pay once at "import time". The function scoped imports are misguided attempts at micro-optimisation imo.

rwalton-arm (changes requested):

src/greentea/__main__.py
```diff
@@ -2,6 +2,6 @@
 # Copyright (c) 2021 Arm Limited and Contributors. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 #
-from mbed_greentea.greentea_cli import main
+from greentea.greentea_cli import main
```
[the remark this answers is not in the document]
wernerlewis: Yes, the README will need to be updated in another PR, both to remove the Mbed references and to bring the content up to date.

src/greentea/gtea/greentea_cli.py
wernerlewis: I've updated to remove any `ctest_` prefixes, as well as code that was previously used when using CTestTestFile.cmake.

src/greentea/gtea/greentea_dlm.py
wernerlewis: I've added a commit to remove this locking implementation, as it doesn't seem to be used except for registering each greentea instance in "kettle.json".

.github/issue_template.md
```diff
@@ -0,0 +1,43 @@
+<!--
```
LDong-Arm, line 1: This template might not be up to date. Maybe we can omit it, and decide what to use in the future?

.github/pull_request_template.md
```diff
@@ -0,0 +1,32 @@
+### Description
```
LDong-Arm, line 1: This template might not be up to date. Maybe we can omit it, and decide what to use in the future?

CONTRIBUTING.md
```diff
@@ -0,0 +1,220 @@
+# Contributing
```
LDong-Arm, line 1: This template might not be up to date. Maybe we can omit it, and decide what to use in the future?

.flake8
```diff
@@ -0,0 +1,15 @@
...
+    test/*,
+ignore =
+    # H301: one import per line
+    H301,
+    # H306: imports not in alphabetical order
+    H306,
+    # W503:  line break before binary operator (this is no longer PEP8 compliant)
+    W503,
+    # E203: whitespace before ':' (this is not PEP8 compliant)
+    E203,
```
LDong-Arm, line 15: It seems to be added by this PR, as it looks different from what exists in mbed-os-tools?
The rules look good to have, maybe we should acknowledge the issues and plan to fix them in the future, instead of ignoring the rules?

LDong-Arm (commented):
Thanks, there are lots of nice improvements in this PR.

But as an initial code-import PR, it might be easier if we just imported mbed-os-tools code "as-is" and make it run? As mbed-os-tools is in a working state, keeping it working in a different repo might only involve updating paths, etc. if I didn't miss anything? Maybe we can enable tests (e.g. tox) and fix failures incrementally in separate PRs?

The commits to import code look good to me, but the other fixes and improvements seem to have some learning curve for reviewers, and there are several things to review. Shall we consider limiting the scope and getting the imported code in first, so subsequent work (incl. other items in #303) can be worked on parallel?

.flake8
wernerlewis: Yes, I added these from mbed-tools. I've removed H301 and H306 as these do not need to be ignored. W503 is ignored as it goes against best practices, and E203 is ignored for PEP compliance, otherwise errors are raised due to formatting of list slicing.

src/greentea/gtea/greentea_hooks.py
```diff
@@ -6,7 +6,8 @@
 import os
 import re
 import json
-from subprocess import Popen, PIPE
+
+from subprocess import run
```
rwalton-arm, line 10: Usually the style is to `import subprocess` and then call `subprocess.run` as it's then explicit what the `run` function is doing. The function name itself is rather generic, so using the namespaced form at the call site is better in this case.

src/greentea/gtea/greentea_hooks.py
```diff
@@ -35,36 +36,27 @@ def __init__(self, name, cmd):
...
-            gt_logger.gt_log_err("hook exited with error: %d, dumping stderr..." % ret)
-            print(_stderr)
-        return ret
+        gt_logger.gt_log_tab(f"hook command: {cmd}")
+        try:
+            p = run(self.cmd, capture_output=True, shell=True)
+        except OSError as e:
+            gt_logger.gt_log_err(str(e))
+            return -1
+        if p.stdout:
```
rwalton-arm, line 47: Why not just set `capture_output=False`? We seem to print `stdout` and `stderr` all the time if they aren't empty.

I would also leave `shell=False` unless absolutely necessary, as `shell=True` causes `subprocess` to spawn an intermediate "shell instance" to run the command, which could make us vulnerable to shell injection exploits.

We might have to change `self.cmd` to a list to make this work with `shell=False`, if it's currently just a string. We should probably deal with the shell spawning in another PR anyway as it was preexisting.

LDong-Arm (approved):
So, I verified using `tox -e py39` and tests passed. Also did a smoke test using `gt`installed from this PR to run tests on a board.

Btw, what about calling the command `greentea` instead of `gt`? I just feel that two-letter commands are too easy to clash, and normally system commands are this short.

@ARMmbed/mbed-os-core I strongly feel that we should prioritise getting this merged, because it's been open for a long time and blocking further work on this repo.

rwalton-arm (changes requested):

rwalton-arm:
> So, I verified using `tox -e py39` and tests passed. Also did a smoke test using `gt`installed from this PR to run tests on a board.
> 
> Btw, what about calling the command `greentea` instead of `gt`? I just feel that two-letter commands are too easy to clash, and normally system commands are this short.
> 
> @ARMmbed/mbed-os-core I strongly feel that we should prioritise getting this merged, because it's been open for a long time and blocking further work on this repo.

I'm OK with the overall commits (apart from one point about how we import `subprocess.run`). One potentially important change is that we've removed the lockfile mechanism when spawning multiple instances. Most of the other changes are just cosmetic or small amounts of refactoring. I'm fine with merging the PR after my latest comment is addressed.

src/greentea/gtea/greentea_hooks.py
wernerlewis: I've removed the `capture_output` arg. `cmd` is a string, so that will need to be changed to remove `shell=True`.

rwalton-arm (approved):
LGTM. Thanks.

LDong-Arm (approved):
LGTM

PR #304, opened 2021-06-07
Update Create call in target_base python

rajkan01:
Access the 'create' via importing from `mbed_ls_tools` instead of detect


rwalton-arm (changes requested):
Please explain in the commit message why we're changing this.

rwalton-arm (changes requested):
Please wrap the body of the commit message at ~72 characters and remove the "-" as it's not a bullet list. You should also aim to keep the summary at 50 characters or less if possible. If the body of the commit message's line length is much longer than 72 characters it becomes quite difficult to read the git log.

rwalton-arm (approved):
LGTM, thanks.

PR #307, opened 2021-06-14
Update the htrun readme

rajkan01:
- Updated the htrun readme based on standalone greentea

### DIFF
--- a/src/htrun/README.md
+++ b/src/htrun/README.md
@@ -1,14 +1,14 @@
 # Development moved
 
-The development of Htrun has been moved into the [mbed-os-tools](../../src/mbed_os_tools) package. You can continue to use this module for legacy reasons, however all further development should be continued in the new package.
+The development of Htrun (mbed-host-tests) has been moved into the [greentea](https://github.com/ARMmbed/greentea/tree/master) repository under `src/htrun`. You can continue to use this module for legacy reasons, however all further development should be continued in the greentea repository.
 
 -------------
 
-[![PyPI version](https://badge.fury.io/py/mbed-host-tests.svg)](https://badge.fury.io/py/mbed-host-tests)
+[![PyPI version](https://badge.fury.io/py/mbed-greentea.svg)](https://badge.fury.io/py/mbed-greentea)
 
-# Htrun (mbed-host-tests)
+# Greentea host tests runner (htrun) command line
 
-`htrun` has extensive command line. In most cases `htrun` (or its command line avatar `mbedhtrun`) will be run in background:
+`htrun` has extensive command line. In most cases `htrun` will be run in background:
 * driving test binary flashing,
 * device reset and
 * test execution.
@@ -29,12 +29,12 @@ Test execution phase will consist of:
 
 ## Command line overview
 
-This chapter will present few examples of how you can use `mbedhtrun` command line to execute tests. In most cases test automation tools such as [Greentea](https://github.com/ARMmbed/greentea) will execute `mbedhtrun` implicitly. There are cases when we want to execute `mbedhtrun` independently. Mostly in situation when we want to:
+This chapter will present a few examples of how you can use `htrun` command line to execute tests. In most cases test automation tools such as [Greentea](https://github.com/ARMmbed/greentea/src/greentea) will execute `htrun` implicitly. There are cases when we want to execute `htrun` independently. Mostly in situation when we want to:
 * debug our code and have binary + host test instrumentation on,
 * prototype or
 * just want to replace test runner in another OS with one compatible with mbed-enabled devices.
 
-All `mbedhtrun` needs is name of the binary you want to flash and method of flashing!
+All `htrun` needs is name of the binary you want to flash and method of flashing!
 
 ## Useful command line end-to-end examples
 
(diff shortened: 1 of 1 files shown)


src/htrun/README.md
```diff
@@ -261,9 +261,9 @@ Key-value protocol was developed and is used to provide communication layer betw
 
 # Greentea client API
```
LDong-Arm, line 262: In my opinion, anything below can be split into a separate file (e.g. programming guide). Also, we could consider moving docs into a `docs/` directory of this repository. This might be outside the scope of the PR and can be done in a follow-up PR.

src/htrun/README.md
```diff
@@ -261,9 +261,9 @@ Key-value protocol was developed and is used to provide communication layer betw
...
 
-DUT test API was first introduced in ```mbedmicro/mbed``` project [here](https://github.com/mbedmicro/mbed/tree/master/libraries/tests/mbed/env). After refactoring this functionality was copied and improved in [greentea-client](https://github.com/ARMmbed/mbed-os/tree/master/features/frameworks/greentea-client) module.
+[greentea-client](https://github.com/ARMmbed/greentea-client) is a C++ client library and provides APIs which can be used to communicate with the Greentea host test tool.
 
-* Slave side key-value protocol API, see [here](https://github.com/ARMmbed/mbed-os/blob/master/features/frameworks/greentea-client/greentea-client/test_env.h) for details.
+* Slave side key-value protocol API, see [here](https://github.com/ARMmbed/greentea-client/blob/master/include/greentea-client/test_env.h) for details.
 
 ```c++
```
LDong-Arm, line 268: I think the two code blocks below can be removed, and the paragraphs can be reworded saying `greentea_send_kv()` sends key-value pairs, `greentea_parse_kv` ... As we already have a link to `test_env.h`, there's no need to have an inline copy of the the API in my opinion, as we would need to keep them updated if anything changes in the future.

src/htrun/README.md
```diff
@@ -481,18 +481,15 @@ int main() {
...
     // Blocking wait for master response for {{gimme_something; some_stuff}}
     greentea_parse_kv(key, value, sizeof(key), sizeof(value));
-    // ...
+    
     fprintf(stderr, "Received from master %s, %s", key, value);
-    // ...
-
     GREENTEA_TESTSUITE_RESULT(true);    // __exit
 }
 ```
```
LDong-Arm, line 495: Snippets could potentially be converted into examples instead in the future.

src/htrun/README.md
```diff
@@ -1,14 +1,14 @@
 # Development moved
 
-The development of Htrun has been moved into the [mbed-os-tools](../../src/mbed_os_tools) package. You can continue to use this module for legacy reasons, however all further development should be continued in the new package.
+The development of Htrun (mbed-host-tests) has been moved into the [mbed-greentea](https://github.com/ARMmbed/greentea/tree/master) repository under `src/htrun`. You can continue to use this module for legacy reasons, however all further development should be continued in the mbed-greentea repository.
```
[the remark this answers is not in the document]
rajkan01: I think this paragraph to give some `warning` to the user as this source moved from a different repo and we will keep this notice until we remove the `htrun` source from `mbed-os-tools`

src/htrun/README.md
rajkan01: Yes once we created the example will update the Readme accordingly

src/htrun/README.md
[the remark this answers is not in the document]
LDong-Arm: Normally a warning is in an old place asking users to look at the new place, rather than in the new place itself like here. I don't mind too much for now as it's a minor detail.

src/htrun/README.md
```diff
@@ -29,12 +29,12 @@ Test execution phase will consist of:
 
 ## Command line overview
 
-This chapter will present few examples of how you can use `mbedhtrun` command line to execute tests. In most cases test automation tools such as [Greentea](https://github.com/ARMmbed/greentea) will execute `mbedhtrun` implicitly. There are cases when we want to execute `mbedhtrun` independently. Mostly in situation when we want to:
+This chapter will present few examples of how you can use `htrun` command line to execute tests. In most cases test automation tools such as [Greentea](https://github.com/ARMmbed/greentea/src/greentea) will execute `htrun` implicitly. There are cases when we want to execute `htrun` independently. Mostly in situation when we want to:
```
LDong-Arm, line 32:
```suggestion
This chapter will present a few examples of how you can use `htrun` command line to execute tests. In most cases test automation tools such as [Greentea](https://github.com/ARMmbed/greentea/src/greentea) will execute `htrun` implicitly. There are cases when we want to execute `htrun` independently. Mostly in situation when we want to:
```

src/htrun/README.md
```diff
@@ -122,91 +122,91 @@ Command line switch format `--fm <config_name>`.
...
 ```
 
-# mbed-host-tests
+# Greentea host tests runner (htrun)
 
-mbed's test suite (codenamed ```Greentea```) supports the *test supervisor* concept. This concept is realized by this module. ```mbed-host-tests``` is a collection of host tests. Host test is script written in Python, which is executed in parallel with the test suite runner (a binary running on the target hardware / device under test) to monitor the test execution's progress or to control the test flow (interaction with the mbed device under test - DUT). The host test is also responsible for grabbing the test result, or deducing it from the test runner's behavior.
+mbed's test suite (codenamed ```Greentea```) supports the *test supervisor* concept. This concept is realized by this module. ```htrun``` is a collection of host tests. Host test is script written in Python, which is executed in parallel with the test suite runner (a binary running on the target hardware / device under test) to monitor the test execution's progress or to control the test flow (interaction with the mbed device under test - DUT). The host test is also responsible for grabbing the test result, or deducing it from the test runner's behavior.
```
LDong-Arm, line 196:
```suggestion
Mbed's test suite (codenamed ```Greentea```) supports the *test supervisor* concept. This concept is realized by this module. ```htrun``` is a collection of host tests. Host test is script written in Python, which is executed in parallel with the test suite runner (a binary running on the target hardware / device under test) to monitor the test execution's progress or to control the test flow (interaction with the mbed device under test - DUT). The host test is also responsible for grabbing the test result, or deducing it from the test runner's behavior.
```

src/htrun/README.md
```diff
@@ -261,31 +261,9 @@ Key-value protocol was developed and is used to provide communication layer betw
...
-
-```c++
-int greentea_parse_kv(char *out_key, char *out_value, const int out_key_len, const int out_value_len);
-```
-
-This function should replace ```scanf()``` used to check for incoming messages from master.
-Function parses input and if key-value message is found load to ```out_key```, ```out_value``` key-value pair. Use ```out_key_size``` and ```out_value_size``` to define out buffers max size (including trailing zero).
+* Slave side key-value protocol API, see [here](https://github.com/ARMmbed/greentea-client/blob/master/include/greentea-client/test_env.h) for details.
 
 # Key-value transport protocol sequence
```
LDong-Arm, line 269: Shall we remove lines 307 to 331 (line numbers in your new version)? [The code has changed](https://github.com/ARMmbed/greentea-client/blob/08fdb8b68f99c523ee298dea6c93eedb60aef9a8/source/greentea_test_env.cpp#L66-L87) so the snippet isn't accurate anymore. Rather than updating the snippet or adding a link, I think we can omit it altogether because diagram there is enough to explain the htrun log.

LDong-Arm (commented):
Thanks @rajkan01. As this PR aims to ensure information in the README is generally correct and up-to-date, without any major refactoring or improvements, it looks good to me.

I've just added a few minor comments when I read through the doc again.

LDong-Arm (approved):
LGTM

PR #321, opened 2021-07-16
Remove gt and cleanup

### LINKED ISSUE #319: Remove `mbedgt`/`gt` from ARMmbed/greentea

rajkan01:
Fixes #319 
-  This PR changes to remove the gt python sources from greentea repo as we decided to use the only htrun for running greentea test using CTest.
- Updated setup.py to release the only htrun
- Moved out htrun Readme from src into root dir


requirements.txt
```diff
@@ -1,6 +1,6 @@
 mbed-ls>=1.8.6
 PySerial>=3.0,<4.0
 setuptools-scm>=1.17.0,<2
-PrettyTable<=1.0.1
+prettytable<3.0,>=2.0
```
rwalton-arm, line 4: Should this be in a separate commit from "Remove gt and cleanup"?

setup.py
```diff
@@ -2,14 +2,14 @@
 # Copyright (c) 2021 Arm Limited and Contributors. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 #
-"""PyPI Package definition for Greentea CLI and host-tests runner (htrun)."""
+"""PyPI Package definition for host-tests runner (htrun)."""
```
rwalton-arm, line 5:
```suggestion
"""PyPI Package definition for greentea-host (htrun)."""
```

setup.py
```diff
@@ -2,14 +2,14 @@
...
 import os
 from io import open
 from distutils.core import setup
 from setuptools import find_packages
 
 DESCRIPTION = (
-    "The Greentea CLI and host-tests runner (htrun) is a collection of tools "
+    "host-tests runner (htrun) is a collection of tools "
     "that enable automated testing on Mbed enabled platforms."
```
rwalton-arm, line 13:
```suggestion
    "greentea-host (htrun) is a command line tool "
    "that enables automated testing on embedded platforms."
```

setup.py
```diff
@@ -53,7 +53,7 @@ def read(fname):
     license="Apache-2.0",
     test_suite="test",
     entry_points={
-        "console_scripts": ["gt=greentea.greentea_cli:main", "htrun=htrun.htrun:main"],
+        "console_scripts": ["htrun=htrun.htrun:main"],
```
rwalton-arm, line 56: Do we want to consider changing the command line to `greentea-host` or do we want to leave it as `htrun`?

rwalton-arm (changes requested):
Nice work. I've left a couple of comments.

One thing I noticed: we seem to have reduced the overall test coverage on the remaining code by 7.81%. Do you have any ideas for how we could increase it back to the previous level?

setup.py
rajkan01: IMO We go with `htrun` as it is actually a host tests runner

rwalton-arm: Fair enough.

rwalton-arm (approved):

